PR #17, opened 2024-03-19
Create Standards

IB-12:


yzueger:
Maybe we could put the standards in a .md file for nice formatting when viewing?
Content looks good to me

joriba:
Agree with yoric, Markdown would be nice

mbouaissi:
Looks good to me, but maybe add more details, with how we should comment. Like saying if we want javadoc, or just one line comments?

Standards
```diff
@@ -0,0 +1,5 @@
+1. One branch per feature
+2. Explicit comments and docs
+3. Clean formatting (KTFMT) (Camel Case)
+4. Use flows or livedata
+5. https://developer.android.com/kotlin/style-guide
```
tsogtb, line 5: Let's include: Avoid hard coded strings and use stringResource whenever possible.

mbouaissi (approved):
lgtm

joriba:
Uhm, why is the last commit empty? @KlaKalma did you forget to `git add` something?

KlaKalma:
Forgot to add sorry, I just renamed, should be good now.

PR #27, opened 2024-03-20
Set up CI similar to the bootcamp

joriba:
I copied the CI file from the bootcamp to the .github/workflows/ folder and changed the names and other properties (such as the VM the workflow runs on) to match the Project. 

For formatting, the CI just runs KTFmt Check to verify that everything is formatted correctly. To automatically format the code before every commit, add a local pre-commit hook: Add the file `dev/hooks/pre-commit` to `<project-directory>/.git/hooks/` and make it executable (on Linux/MacOS: `chmod +x .git/hooks/pre-commit`. 
NOTE: the hook file can cause nasty merge conflicts when only parts of a file are commited and not yet properly formatted. Either only commit whole files or make sure to run KTFmt first (manually).

Tsogt also created a new Sonarcloud workflow for static code analysis


.github/workflows/CI.yaml
```diff
@@ -0,0 +1,102 @@
...
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: gradle-${{ runner.os }}-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+
+      # To Load a fake google map api key for the test 
+      - name: Loading Local.properties (dummy)
+        run: |
+          echo MAPS_API_KEY=\"0\" > ./local.properties
```
yzueger, line 63: I would suggest directly referencing secrets, seems pretty straight forward:

https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow

.github/workflows/CI.yaml
```diff
@@ -0,0 +1,102 @@
...
+          avd-name: github
+          force-avd-creation: true
+          emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
+          disable-animations: true
+          script: ./gradlew connectedCheck --parallel --build-cache
+
+      # This step generates the coverage report which will be used later in the semster for monitoring purposes
+      #- name: Generate coverage
+      #  run: |
+      #    ./gradlew jacocoTestReport
```
yzueger, line 102: Would be nice to get the report. This might work:
https://github.com/marketplace/actions/jacoco-report

yzueger (commented):
Some impreovement suggestions in comments

joriba:
The check fails because of insufficient test coverage. Writing and testing code is not really the purpose of this PR, so I would merge anyway...

yzueger:
Shouldn't there be a comment with the Jacoco coverage?

But also with current branch protection I think it gets blocked from merging when the CI fails.

joriba:
> Shouldn't there be a comment with the Jacoco coverage?
What do you mean, a comment? There is one just above...

> But also with current branch protection I think it gets blocked from merging when the CI fails.
I don't think so. When the PR is reviewed and approved, the merge button will be gray, but still clickable from what I have seen.

yzueger (approved):
My bad, I didn't realize the Jacoco report edits the message instead of creating a new one after newer commits. With that seemingly working, lgtm

PR #46, opened 2024-03-22
Integrate sonarcloud into the main CI

joriba:
I integrated the SonarCloud workflow into the main CI and deleted the old one. DRAFT, so it might not work yet :)


yzueger (approved):
Looks like it's working according to #48, lgtm

joriba:
As the test branch has shown, this should now be ready to merge. Test coverage is only found if there actually are tests, which is why this hoop was necessary, but after merging this as well as the app skeleton, we should have CI coverage report on the main branch.

PR #50, opened 2024-03-24
32 display a map

joriba:
I added a simple GoogleMap component that displays a navigatable google map as one of the destinations in the bottom app bar. 
At this point, it's not very sophisticated, but I suggest already merging this (working) version into main, so it can be integrated with the beacons etc.


tsogtb:
> In the future, the map should probably be the default first view, and
(depending on UI choices) fill the whole screen (?) 

I agree that no bottom bar on map screen will give the user a better feel.
If we navigate by swiping as was illustrated in Figma, we would probably have to allocate a thin portion of the screen on the sides or track the swiping speed (?) 
I am thinking about a workaround where we toggle the bottom bar on click/drag actions so that when the user is actively searching for beacons its full-screen.

KlaKalma (approved):
lgtm

Issue #40, opened 2024-03-22
Have the possibility to choose if my profile is public or private to others users


aymanbakiri:

mbouaissi:
Will be done by thursday

PR #55, opened 2024-03-26
Profile

mbouaissi:
Hello team, 

This is what is implemented in this pull request:

- [ ] #40 
- [ ] #37 (not entirely done, but a todo is line 306 to indicate to put here the logic to sending information to the db)

Summary of what has been done:

- ViewModel of the profile
- A profile screen where the user can see his profile
- A edit view, where the user can modify his profile (first name, last name, picture and description)

(# of hours on this 10)


app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/ProfileScreen.kt
```diff
@@ -0,0 +1,406 @@
...
+ *
+ * @author Menzo Bouaissi
+ * @since 1.0
+ * @last update 1.0
+ */
+@Preview
+@Composable
+fun ProfileScreenPreview() {
+  ProfileScreen(ProfileViewModel())
+}
```
tsogtb, line 406: I think from the maintainability perspective it's better to extract the components out. I really like the documentation and we should adopt the habit as a team.

app/src/main/java/ch/epfl/cs311/wanderwave/model/data/Profile.kt
```diff
@@ -0,0 +1,12 @@
...
+import android.net.Uri
+
+data class Profile(
+    var firstName: String,
+    var lastName: String,
+    var description: String,
+    var numberOfLikes: Int,
+    var isPublic: Boolean,
+    var profilePictureUri: Uri? = null
+)
```
tsogtb, line 12: Should we include a field for a currently "on repeat" song. It could be like a user status. We can maybe restrict this feature so that people should first collect points to unlock?

tsogtb: Speaking of points, I think we can also add the field for the currency.

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/ProfileScreen.kt
```diff
@@ -0,0 +1,406 @@
...
+ * This is the screen composable, which can either show the profile of the user and the chosen
+ * songs, or it could show a view to modify the profile
+ *
+ * @param viewModel the viewModel that will handle the profile *
+ * @author Menzo Bouaissi
+ * @since 1.0
+ * @last update 1.0
+ */
+@Composable
+fun ProfileScreen(viewModel: ProfileViewModel = hiltViewModel()) {
```
tsogtb, line 71: Don't think passing it as a parameter is necessary here.

app/build.gradle.kts
```diff
@@ -125,6 +122,12 @@ dependencies {
     androidTestImplementation("io.mockk:mockk:1.13.10")
     androidTestImplementation("io.mockk:mockk-android:1.13.10")
     androidTestImplementation("io.mockk:mockk-agent:1.13.10")
+
+    // Dependencies for the photo part
+    implementation("io.coil-kt:coil-compose:2.6.0")
+
+    // Dependencies for the Observable part
+    implementation("androidx.compose.runtime:runtime-livedata:1.0.0-alpha07")
```
KlaKalma, line 130: Didn't we say we would use flows ?

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/ProfileViewModel.kt
```diff
@@ -0,0 +1,37 @@
+package ch.epfl.cs311.wanderwave.viewmodel
+
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.ViewModel
+import ch.epfl.cs311.wanderwave.model.data.Profile
+
+class ProfileViewModel : ViewModel() {
+  private val _profile =
```
KlaKalma, line 9: Use flows

KlaKalma (changes requested):
I thought we would use flow ? I think we can still discuss that as it is already tested and implemented, but would be nice to be consistent

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/ProfileScreen.kt
tsogtb: No I meant for the overall function I think it is better to define new files inside ui.components.profile and make the Profile screen shorter. Sorry for the confusion.

KlaKalma (approved):
lgtm, may need another review because it's too much

PR #57, opened 2024-03-27
Handle bottom bar visibility explicitly

yzueger:
This PR refactors the navigation to make it explicit which screens show the bottom bar and which ones don't.
It also removes some references to TOP_LEVEL_DESTINATIONS for usages that could just use a route directly.


app/src/androidTest/java/ch/epfl/cs311/wanderwave/ui/AppBottomBarTest.kt
mbouaissi, line 1: Do we really need to pass Route.MAIN.routeString? Otherwise It's much clearer

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/AppBottomBar.kt
mbouaissi, line 1: Same as before, but do we really need currentRoute?

tsogtb (approved):
I think it works out nicely because even though it was excluded inside the screens, for some reason when the app launches the bottom bar was showing for a split second.

mbouaissi (approved):
Overall, the code is much nicer and cleaner and the only small issue I see is the currentRoute string.

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/AppBottomBar.kt
yzueger: No, that was left over from the previous implementation. We would need it if we decide to highlight the current route though.

joriba (approved):
Lgtm

KlaKalma:
lgtm

KlaKalma (approved):

PR #61, opened 2024-03-27
Exclude ui from Sonar and move navigation out of ui package

yzueger:
Actual time: 1h


KlaKalma (approved):
lgtm, thanks

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/DummyScreen.kt
joriba, line 1: Do we still need this?

joriba (approved):
Do we still need the DummyScreen? 
Otherwise lgtm

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/DummyScreen.kt
yzueger: No we don't, but I didn't want to delay the PR. I don't think deleting it later will cause merge conflicts

PR #63, opened 2024-03-27
Login Screen enhancement

### LINKED ISSUE #80: Create an Info Screen

tsogtb:
Just a follow up on the Login Screen, added AboutScreen for future use.  
revisits #41 
Includes a bug fix.
https://github.com/G12-Wanderwave/Wanderwave/assets/84101643/bfbdfbad-deb3-4dee-831e-273857e031a5
also,
closes #80


app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/about/AboutHeader.kt
```diff
@@ -0,0 +1,25 @@
...
+@Composable
+fun AboutHeader(modifier: Modifier) {
+  Column(
+      horizontalAlignment = Alignment.CenterHorizontally,
+      verticalArrangement = Arrangement.Center,
+      modifier = modifier.fillMaxWidth()) {
+        Text(text = "Wanderwave", style = MaterialTheme.typography.displayMedium)
+        Text(
+            text = "alpha v1.0",
+            style = MaterialTheme.typography.titleMedium,
```
yzueger, line 22: Don't hardcode strings, we should be able to use `BuildConfig.VERSION_NAME`

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/animated/AnimatedIcon.kt
```diff
@@ -0,0 +1,53 @@
...
+import androidx.compose.runtime.getValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.unit.dp
+import ch.epfl.cs311.wanderwave.R
+import ch.epfl.cs311.wanderwave.ui.theme.placeholderColor
+
+@Composable
+fun AnimatedIcon() {
```
yzueger, line 25: Maybe reference in the name that this is the app logo icon

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/login/LoginScreenHeader.kt
```diff
@@ -2,34 +2,69 @@ package ch.epfl.cs311.wanderwave.ui.components.login
...
-                Modifier.height(50.dp)
-                    .fillMaxWidth()
-                    .padding(start = 50.dp, end = 50.dp)
-                    .testTag("appLogo"))
+fun LoginScreenHeader(navigationActions: NavigationActions, modifier: Modifier) {
+  val startColor = MaterialTheme.colorScheme.primary
+  val endColor = placeholderColor
+  val colorSpots = List(10) { i -> lerp(startColor, endColor, i / 9f) }
+  FloatingActionButton(
+      onClick = { navigationActions.navigateTo(Route.ABOUT) },
```
yzueger, line 34: I don't think the Login Screen is the best place to link to the about screen. From my experience, it's usually in the settings or in the users profile where it's linked. The login screen will be seen for a very short time before pressing the login button by most users.
If we leave it here, agian, FloatingActionButton seems weird.

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/login/LoginScreenHeader.kt
```diff
@@ -2,34 +2,69 @@ package ch.epfl.cs311.wanderwave.ui.components.login
...
+              modifier = Modifier.fillMaxWidth()) {
+                Icon(
+                    painter = painterResource(id = R.drawable.wanderwave_icon),
+                    contentDescription = "app icon",
+                    tint = colorSpots[0],
+                    modifier = Modifier.height(40.dp).testTag("appIcon"))
+                val anderwave: List<String> = listOf("a", "n", "d", "e", "r", "w", "a", "v", "e")
+                anderwave.forEachIndexed { index, letter ->
+                  FillLogo(letter = letter, colorSpot = index, colorSpots = colorSpots)
+                }
```
yzueger, line 57: If that's how we draw the logo, we should probably extract it to a reusable component. It would probably be easiest to just use something like an SVG instead.

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/login/SignInButton.kt
```diff
@@ -18,19 +19,23 @@ import androidx.compose.ui.res.painterResource
...
-          modifier = Modifier.height(27.dp))
-      Spacer(modifier = Modifier.width(12.dp))
-      Text(
-          text = stringResource(id = R.string.button_label),
-          style = MaterialTheme.typography.titleMedium)
-    }
+    Button(
+        onClick = { onClick() },
+        modifier = Modifier.testTag("signInButton"),
+        colors = ButtonDefaults.buttonColors(containerColor = signInButtonColorPlaceHolder)) {
```
yzueger: I would probably use Spotify green as button color, but it doesn't really matter if it's just a placeholder

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/login/WelcomeTitle.kt
```diff
@@ -2,39 +2,88 @@ package ch.epfl.cs311.wanderwave.ui.components.login
...
+                  painter = painterResource(id = R.drawable.hexagon),
+                  contentDescription = stringResource(id = R.string.circle_icon),
+                  tint = Color.Blue)
+              LoginScreenPulse(startColor = Color.Blue, endColor = Color.Magenta, reverse = false)
+              Icon(
+                  painter = painterResource(id = R.drawable.hexagon),
+                  contentDescription = stringResource(id = R.string.circle_icon),
+                  tint = Color.Blue,
+                  modifier = Modifier.size(20.dp))
+            }
```
yzueger, line 83: While these animated lines are fancy, I don't really see the point of having them on the login screen. From a UX perspective, the user should be directed to press the login button and not really do much else on the login screen.

app/src/main/java/ch/epfl/cs311/wanderwave/ui/theme/Color.kt
```diff
@@ -2,6 +2,8 @@ package ch.epfl.cs311.wanderwave.ui.theme
 
 import androidx.compose.ui.graphics.Color
 
+val placeholderColor = Color(0xFFFF0000)
+val signInButtonColorPlaceHolder = Color(0xFFFFA500)
```
yzueger: I don't think we should put individual component colors in the global theme, either use one of the reusable theme colors, or just define the color in the component itself

yzueger (dismissed):

app/src/main/java/ch/epfl/cs311/wanderwave/navigation/NavigationActions.kt
```diff
@@ -59,6 +67,13 @@ class NavigationActions(navController: NavHostController) {
   fun goBack() {
     navigationController.popBackStack()
   }
+
+  fun signIn() {
```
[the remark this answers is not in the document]
tsogtb: If we just navigateToTopLevel the stack isn't cleared so user can manually go back to Login. The only possible way to navigate to Login must be through sign out button.

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/login/LoginScreenHeader.kt
tsogtb: Agree. The way it's handled now is a bit clumsy

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/login/LoginScreenHeader.kt
tsogtb: Agree that simpler design is better.

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/login/SignInButton.kt
tsogtb: I tried Spotify green, not sure if I like it or not. I think we need a team decision.

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/login/WelcomeTitle.kt
tsogtb: Agreed.

app/src/main/java/ch/epfl/cs311/wanderwave/navigation/NavigationActions.kt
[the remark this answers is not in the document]
yzueger: Shouldn't that be the behaviour when navigating to all topLevelDestionations? It indeed looks like it doesn't behave like this at the moment, but I would consider that a bug of the `navigateToTopLevel` implementation.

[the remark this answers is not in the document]
tsogtb: Yeah I am not so sure for the moment, maybe we should consider refactoring navigation. And I wanted this behavior because after signing in and presented on the mainscreen when user clicks back button of the phone the app should minimize instead of navigating back because mainscreen is at the index 0 of the navigation stack after logging in. I think it behaves like so for the moment.

app/src/main/java/ch/epfl/cs311/wanderwave/navigation/NavigationActions.kt
```diff
@@ -29,6 +30,18 @@ class NavigationActions(navController: NavHostController) {
   private var _currentRouteFlow = MutableStateFlow(getCurrentRoute())
   val currentRouteFlow: StateFlow<Route?> = _currentRouteFlow
 
+  // Handle user manually clicking the back button
+  init {
+    navController.addOnDestinationChangedListener { _, destination, _ ->
```
mbouaissi, line 35: This part is fine, but I think you can use .let to have all in a line:
`      _currentRouteFlow.value = destination.route?.let { Route.forRouteString(it) }
`

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/login/LoginAppLogo.kt
```diff
@@ -0,0 +1,56 @@
...
+              val anderwave: List<String> = listOf("a", "n", "d", "e", "r", "w", "a", "v", "e")
+              anderwave.forEachIndexed { index, letter ->
+                FillLogo(letter = letter, colorSpot = index, colorSpots = colorSpots)
+              }
+            }
+      }
+}
+
+@Composable
+private fun FillLogo(letter: String, colorSpot: Int, colorSpots: List<Color>) {
```
mbouaissi, line 50: Can't we just pass a Color instead of a index and a list?

mbouaissi (approved):
Overall, the animation look sick, good job! I have just left some comments about minor things. Also maybe try to add some comments on the different functions, but maybe about the comments, we should discuss as a team how we want to do it. Otherwise great job

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/login/LoginAppLogo.kt
tsogtb: True

PR #65, opened 2024-03-27
Authenticate to on-device Spotify app

### LINKED ISSUE #43: Prompt user to Spotify sign-in flow upon clicking "Sign-in with Spotify"
Estimated time: 6 hours
Actual time (Sprint 2): 13 hours

yzueger:
This PR adds an authentication flow to connect to the on-device Spotify app. It can be controlled via the newly created SpotifyController.

Closes #43


app/src/main/java/ch/epfl/cs311/wanderwave/ui/App.kt
```diff
@@ -53,14 +64,17 @@ fun AppScaffold(navController: NavHostController) {
...
-            startDestination = Route.LOGIN.routeString,
+            startDestination = Route.SPOTIFY_CONNECT.routeString,
             modifier = Modifier.padding(innerPadding)) {
-              composable(Route.LOGIN.routeString) { LoginScreen(navActions) }
+              composable(Route.SPOTIFY_CONNECT.routeString) { SpotifyConnectScreen(navActions) }
+              composable(Route.LOGIN.routeString) { LoginScreen(navActions, showSnackbar) }
               composable(Route.MAIN.routeString) { MainPlaceHolder(navActions) }
-              composable(Route.TRACK_LIST.routeString) { TrackListScreen() }
+              composable(Route.TRACK_LIST.routeString) { TrackListScreen(showSnackbar) }
+              composable(Route.TRACK_LIST.routeString) { TrackListScreen(showSnackbar) }
```
tsogtb, line 77: Duplicate here

app/src/main/java/ch/epfl/cs311/wanderwave/ui/App.kt
```diff
@@ -41,8 +46,14 @@ fun App(navController: NavHostController) {
 fun AppScaffold(navController: NavHostController) {
   val navActions = NavigationActions(navController)
   var showBottomBar by remember { mutableStateOf(false) }
-
   val currentRouteState by navActions.currentRouteFlow.collectAsStateWithLifecycle()
+  val snackbarHostState = remember { SnackbarHostState() }
+
```
tsogtb, line 51: I guess it doesn't matter that much for now but Snackbar in middle would be nicer instead of on the bottom which is right on top of the bottom bar.

app/src/main/java/ch/epfl/cs311/wanderwave/model/spotify/SpotifyController.kt
```diff
@@ -0,0 +1,107 @@
...
+      awaitClose { callResult?.cancel() }
+    }
+  }
+
+  enum class ConnectResult {
+    SUCCESS,
+    NOT_LOGGED_IN,
+    FAILED
+  }
+}
```
tsogtb, line 107: nice

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/SpotifyConnectScreenViewModel.kt
```diff
@@ -0,0 +1,32 @@
...
+import androidx.lifecycle.ViewModel
+import ch.epfl.cs311.wanderwave.model.spotify.SpotifyController
+import dagger.hilt.android.lifecycle.HiltViewModel
+import javax.inject.Inject
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.single
+
+@HiltViewModel
+class SpotifyConnectScreenViewModel
```
tsogtb, line 12: nicely done

tsogtb (approved):
Looks neat.

tsogtb:
Hi I merged your branch into mine and resolved the conflicts but for some reason i commited my changes to yours unknowingly, I'm a git newbie I hope you don't mind.

app/src/main/java/ch/epfl/cs311/wanderwave/navigation/NavigationActions.kt
```diff
@@ -29,6 +31,18 @@ class NavigationActions(navController: NavHostController) {
   private var _currentRouteFlow = MutableStateFlow(getCurrentRoute())
   val currentRouteFlow: StateFlow<Route?> = _currentRouteFlow
 
+  // Handle user manually clicking the back button
```
mbouaissi, line 34: I think it's the same part as an other reviewm but i will still add the same comment :
This part is fine, but I think you can use .let to have all in a line:
     _currentRouteFlow.value = destination.route?.let { Route.forRouteString(it) }

mbouaissi (approved):
Good job! There is a lot of code but it was nicely done! I think, maybe we could add a little more comments, but otherwise it's nice. Also, I just have a concern about how you handled the hiltViewModel, because in my PR (https://github.com/G12-Wanderwave/Wanderwave/pull/81), I was advised to not pass the hiltViewModel as an argument, but rather instantiate it where we need it. So I don't know if it's necessary for you to have it as an argument, but if not, maybe try to instantiate it inside your functions

app/src/main/java/ch/epfl/cs311/wanderwave/ui/App.kt
yzueger: I think in Material 3, Snackbars are intended to be at the bottom for small messages that persist across different screens. For important modal dialogs it's probably better to put them directly in the screens, as the user won't be navigating while they are shown anyway.

yzueger:
> Good job! There is a lot of code but it was nicely done! I think, maybe we could add a little more comments, but otherwise it's nice. Also, I just have a concern about how you handled the hiltViewModel, because in my PR (#81), I was advised to not pass the hiltViewModel as an argument, but rather instantiate it where we need it. So I don't know if it's necessary for you to have it as an argument, but if not, maybe try to instantiate it inside your functions

We need to pass it as parameter so we can use a mock viewmodel for UI unit tests. I'm not sure why we wouldn't want to have the viewModel as parameter with a default value, as the only thing it changes is that it makes it impossible to pass a different viewModel.

yzueger:
> Hi I merged your branch into mine and resolved the conflicts but for some reason i commited my changes to yours unknowingly, I'm a git newbie I hope you don't mind.

It seems the commits are all mixed together here, I'll try to force push the state before the merge to fix it

app/src/main/java/ch/epfl/cs311/wanderwave/navigation/NavigationActions.kt
IB-12: He made the change so I'll mark this as resolved

PR #81, opened 2024-03-29
Profile edit

mbouaissi:
The profile and the edit view are ready


app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/ProfileScreen.kt
```diff
@@ -68,11 +67,10 @@ val INPUT_BOX_NAM_SIZE = 150.dp
  * @last update 1.0
  */
 @Composable
-fun ProfileScreen(viewModel: ProfileViewModel = hiltViewModel()) {
+fun ProfileScreen(viewModel: ProfileViewModel) {
```
tsogtb, line 72: Here, I meant refactor
``
@Composable
fun ProfileScreen() {
  val viewModel: ProfileViewModel = hiltViewModel()
  /**/
}
``
like so.

mbouaissi: Oh I see now, but the thing is that maybe later, we will connect few things that are based on the profile, like the authentication screen, ... , and I thought it was better to give it as an argument.
Because if we are using the profileViewmodel in other classes, can we still share the data without passing it as an argument?

tsogtb: I think we are able to do so. we can just do 
`val viewModel: ProfileViewModel = hiltViewModel()` in the places that need it
and hilt will take care of the rest.

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/ProfileScreen.kt
```diff
@@ -0,0 +1,398 @@
...
+) {
+
+  IconButton(
+      modifier = modifier.then(Modifier.size(24.dp)).testTag("clickableIcon"),
+      onClick = { viewModel.toggleEditMode() }) {
+        Icon(
+            icon,
+            contentDescription = "Edit",
+            modifier = modifier,
+            // tint = if (isInEditMode) MaterialTheme.colorScheme.onSurface else
```
tsogtb, line 152: We agreed on getting rid of commented out codes and just put it back later on if we need it since github has the history.

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/ProfileScreen.kt
```diff
@@ -0,0 +1,398 @@
...
+ * Display a visit card with the information of the user
+ *
+ * @param modifier the modifier to be applied to the Icon
+ * @param profile the profile of the user
+ * @author Menzo Bouaissi
+ * @since 1.0
+ * @last update 1.0
+ */
+@Composable
+fun VisitCard(modifier: Modifier = Modifier, profile: Profile) {
```
tsogtb, line 168: It'd be nice to have this in a separate file inside components

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/ProfileScreen.kt
```diff
@@ -0,0 +1,398 @@
...
+        contentDescription = "Profile picture",
+        modifier = modifier)
+  } else {
+    Image(
+        painter = painterResource(id = R.drawable.profile_picture),
+        contentDescription = "Profile picture",
+        modifier = modifier)
+  }
+}
+
```
tsogtb, line 385: over 300 lines of code in one place seems a bit hard to maintain when the project gets bigger. Let's break this down a little bit maybe?

tsogtb (approved):
Let's just get rid of the conflicts and maybe consider the suggestions otherwise looks great to me.

mbouaissi:
@tsogtb thank you very much for your review, I have try to comply with your suggestions, I hope I haven't forgot anything

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/ImagesActions.kt
```diff
@@ -0,0 +1,80 @@
...
+@Composable
+fun ImageSelection(profile: Profile, onImageChange: (Uri?) -> Unit) {
+  var imageUri by remember { mutableStateOf<Uri?>(null) }
+  val launcher =
+      rememberLauncherForActivityResult(contract = ActivityResultContracts.GetContent()) { uri: Uri?
+        ->
+        imageUri = uri
+        onImageChange(uri)
+      }
+  if (imageUri != null) profile.copy(profilePictureUri = imageUri)
```
IB-12, line 69: Couldn't we move the if at line 69 into the launcher above ? We could get rid of the imageUri var that way.
```suggestion
  val launcher =
      rememberLauncherForActivityResult(contract = ActivityResultContracts.GetContent()) { uri: Uri?
        ->
        onImageChange(uri)
        if (uri!= null) profile.copy(profilePictureUri = uri)
      }
```

mbouaissi: done

IB-12 (approved):
lgtm 👍

Issue #77, opened 2024-03-28
Media player can play music


yzueger:
- [x] #114 
- [x] #123 
- [x] #117 
- [x] #128 
- [x] #129 
- [x] #132
- [x] #110 
- [x] #65

PR #84, opened 2024-04-03
77 play music UI skeleton

tsogtb:
builds towards #77 
- Implements the basic components needed for the extendable player UI. 
- As of now, the bottom sheet peeks at **144.dp** (BottomNavBar height included) Although it is made slightly transparent, it's not recommended to display crucial ui components at the bottom of the screens (unless there is a bottom padding).


tsogtb:
https://github.com/G12-Wanderwave/Wanderwave/assets/84101643/55cd3ec7-ad6d-48b9-805b-a180df6d3f59

Here is a demo.

IB-12:
That demo looks good ! Maybe we could have the blurred album cover as a background for the song title ? And it would be nice to make sure we add usual spotify controls (next, previous, add to favorites).

tsogtb:
> That demo looks good ! Maybe we could have the blurred album cover as a background for the song title ? And it would be nice to make sure we add usual spotify controls (next, previous, add to favorites).

Thanks for the suggestion! I totally agree that we should have a matching background. I am also thinking we can make it slightly transparent so that it doesn't disturb the user in map screen as much.

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/MiniPlayer.kt
```diff
@@ -0,0 +1,78 @@
...
+                } else {
+                  onPauseClick()
+                }
+              }) {
+                Icon(
+                    painter =
+                        if (isPlaying) painterResource(id = R.drawable.pause_icon)
+                        else painterResource(id = R.drawable.play_icon),
+                    contentDescription = "Play Pause Icons",
+                    tint = if (isPlaying) Color(0xFF1DB954) else MaterialTheme.colorScheme.primary,
```
IB-12, line 69: I think it might be beneficial to have all colors defined in a specific file (as was done for the bootcamp), to make sure we can propagate changes quickly and easily

tsogtb:
> az_recorder_20240403_161132.mp4
> 
> Here is a demo.

Outdated.

mbouaissi (commented):
I couldn't review yet, because when I checkout on this branch, they were some dependencies missing. Is it a problem on my side, or did you forgot to push the build.gradle?

tsogtb:
> I couldn't review yet, because when I checkout on this branch, they were some dependencies missing. Is it a problem on my side, or did you forgot to push the build.gradle?

Normally it should build since it passed the workflow, I would try syncing Gradle or invalidating the caches and restarting the IDE.

mbouaissi (dismissed):
(First of, sorry, I didn't first go to the correct branch, I went on 77-play-music instead of 77-play-music-ui, that's why the gradle was missing dependencies :( )
It's really impressive, and the ui look very cool! I really like the fact that when I click on the title I will actually see a picture of the album. I just have a few suggestions:

1. Maybe, instead of having 1000L and 500L we could have constant like ONE_SECOND or  HALF_SECOND(but maybe it's not relevant if we plan to have exotic time variation)
2. One thing that could be cool, is when I click on the player, I see the album etc,..., but maybe, when I click to the navigation bar, the player go back to a small size. (Example: I am on the main screen, I click on the player, I see the player in big, then I click on track list, the player get smaller so I can see the track list )
3. I think also that when I have the player in big screen I should be able to pause/continue the music.

Overall, it's very nice what you did, and all the things that I have said are suggestions for the future, we could maybe create issues about it once your PR is merged 👍

tsogtb:
https://github.com/G12-Wanderwave/Wanderwave/assets/84101643/5956cb88-7faf-4642-9dad-9f16c32ab267

Here is what it looks like for the moment, I figured we can split it up into 2 or 3 PRs since the player needs buttons like adding to a beacon and blacklisting etc. 
One more thing that needs improvement in the future is smooth transitions, i.e., loading screen. (If you've noticed, Spotify does so with the combination of green circular indicator, black screen and transparent frozen screen)

Apparently I need to refactor the code some more because SonarCloud says the cognitive complexity is too much.

Lastly, I am not too sure about the animation since we are aiming for nice and simple UI so I would appreciate comments on that.

app/build.gradle.kts
```diff
@@ -13,9 +13,6 @@ plugins {
 
     // handling secrets.properties
     id("com.google.android.libraries.mapsplatform.secrets-gradle-plugin")
-
-    // google firebase
-    id("com.google.gms.google-services")
 }
```
IB-12, line 16: Was there any particular reason for removing the firebase dependencies ?

IB-12 (approved):
Really nice job, it's great to see our app coming to life ! :) I like how you cleaned up the code since the last time I've seen it, makes it easier to read.
Regarding animations, it's kind of hard to judge what is working and what's not so far, I personally feel like we should focus on having the app be fully functional (which should be separate PRs, as you've suggested) and then tie everything together with a nice theme and cool animations. But what you've done so far is very interesting, it shows there are a lot of possibilities available to us. Overall, great job !

app/build.gradle.kts
tsogtb: Sorry for the confusion, I wasn't aware of the latest commits in main and messed up the branch. With the latest commit things should be okay.

mbouaissi (approved):
lgtm :+1:

PR #85, opened 2024-04-05
add secrets.properties and *.keystore to gitignore

joriba:
They were removed when some merge was reverted. Put it into a separate PR now, so it is more robust.


mbouaissi (approved):
lgtm

PR #88, opened 2024-04-06
Show location on map

joriba:
Displays the user's location on the map, using GoogleMap's isLocationEnabled property.
If the user does not grant access to the fine location, don't display it.


app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/MapScreen.kt
```diff
@@ -1,13 +1,39 @@
...
+              Manifest.permission.ACCESS_COARSE_LOCATION,
+              Manifest.permission.ACCESS_FINE_LOCATION,
+          ))
+
+  LaunchedEffect(key1 = permissionState) {
+    // Filter permissions that need to be requested.
+    val permissionsToRequest = permissionState.permissions.filter { !it.status.isGranted }
+
+    // If there are permissions to request, launch the permission request.
+    if (permissionsToRequest.isNotEmpty()) permissionState.launchMultiplePermissionRequest()
```
IB-12, line 31: Could be nice to have a small pop-up message letting the user know that the app relies on GPS access to function if they deny the permission ?

joriba: Good Idea. I'll do something with a dialog box.

joriba:
I'm not sure why, but when merging with main, it seemed as though you guys want to remove the `canNavigateTo..` tests in `AppTest.kt`. Why is that?

yzueger:
This is because the app now connects to Spotify before it reaches the main screens with the bottom bar. If Spotify is not installed, like in the emulator locally and in the CI, this simply won't work.

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/MapScreen.kt
```diff
@@ -1,13 +1,121 @@
...
+  return location?.let { LatLng(it.latitude, it.longitude) }
+}
+
+@OptIn(ExperimentalPermissionsApi::class)
+@SuppressLint("MissingPermission")
 @Composable
-@Preview
-fun MapScreen() {
-  GoogleMap(modifier = Modifier.testTag("mapScreen"))
+fun MapScreen(navigationActions: NavigationActions, viewModel: MapViewModel = hiltViewModel()) {
```
mbouaissi, line 64: Why do we need navigations actions now?

mbouaissi (approved):
For now, at least on my phone it works fine :+1:. I also think that it's nice coded (just maybe try to modularize MapScreen ;) ). Also, I think for now it resolves nicely the issue, as it shows the last location of the user on the map, but what we could do for the next sprint is having an issue to show the live location on the map to. 
But yeah good job

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/MapScreen.kt
joriba: We don't need it yet, but we will definitely need it in the future, to make the Beacons clickable. I already put it in for that.

joriba:
@mbouaissi I'm re-requesting your review because I had to change a lot in your code from your last PR to make it compatible with mine.
Turns out you were using the "old" googleMap API (I think?) We don't actually need the GoogleMap class, just the Composable. I integrated your work into mine, please check that I did not remove anything important or fucked something else up.

joriba:
Second note: the location should now update on the map screen, although still not with the desired frequency, and it still is not very precise. So if anyone has an idea what could be going on, any input will be appreciated :)

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/MapScreen.kt
```diff
@@ -1,79 +1,163 @@
...
+    LaunchedEffect(location != null, mapIsLoaded.value) {
+      if (location != null && mapIsLoaded.value) {
+        moveCamera(cameraPositionState, location, viewModel.cameraPosition.value)
       }
     }
   }
 }
 
+@Composable
+fun WanderwaveGoogleMap(cameraPositionState: CameraPositionState, viewModel: MapViewModel) {
```
KlaKalma, line 91: can't you prevent copying of code by calling this function in the other overloaded function ? or vice versa ? 
`fun WanderwaveGoogleMap(cameraPositionState: CameraPositionState, viewModel: MapViewModel) {
WanderwaveGoogleMap(
    cameraPositionState: CameraPositionState = cameraPositionState,
    viewModel: MapViewModel = viewModel,
    onMapLoaded: () -> Unit = {},
)
`

joriba: I agree that the duplication is a bit unfortunate... the problem is that only the second implementation (with the onMapLoaded function) should set isMyLocationEnabled to true.

I guess it's better to only have one function and pass isLocationEnabled as an argument as well.

joriba:
@KlaKalma better? :)

KlaKalma (approved):
Perfect :)

PR #89, opened 2024-04-07
86 map UI

mbouaissi:
The map UI is done. Unfortunately, I couldn't use GoogleMap Plateform because it uses fragments, and I try to implement it, but the ID of the map didn't work for some reason. Instead, I have used https://mapstyle.withgoogle.com/, that generate a JSON of a map. I'm not a designer, so please tell me if some part of the map are not good, and I will do my best to improve it!


KlaKalma (changes requested):
The UI looks great !

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/MapScreen.kt
```diff
@@ -1,13 +1,94 @@
...
+    } else {
+      googleMapState?.let { googleMap ->
+        val epfl = LatLng(46.518831258, 6.559331096)
+        googleMap.addMarker(MarkerOptions().position(epfl).title("Marker at EPFL"))
+        googleMap.moveCamera(CameraUpdateFactory.newLatLng(epfl))
+      }
+    }
+  }
 }
+
```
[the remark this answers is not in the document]
mbouaissi: I let it in case someone really wanted to use the Google Maps, he would not have to re-write the code again. But once the PR will be reviewed, I will delete it

app/src/main/res/layout/fragment_map.xml
```diff
@@ -0,0 +1,7 @@
+<fragment xmlns:android="http://schemas.android.com/apk/res/android"
```
IB-12, line 1: Is there any particular reason for using Fragments instead of Composables ? I think it would be better to keep it uniform if possible, but it's no big deal.

IB-12 (approved):
Very good job overall. I just had a small question about the use of fragments but it's not critical. Your code is clean and easy to understand, so thanks for making my job easier !

app/src/main/res/layout/fragment_map.xml
mbouaissi: I have just put there in case in the future one would really wants to use google maps plateform, but I can remove them

mbouaissi (commented):
Changes are done

mbouaissi (commented):
:+1:

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/MapScreen.kt
```diff
@@ -1,13 +1,66 @@
...
+    if (!isMapReady.value) {
+      CircularProgressIndicator(modifier = Modifier.testTag("CircularProgressIndicator"))
+    } else {
+      googleMapState?.let { googleMap ->
+        val epfl = LatLng(46.518831258, 6.559331096)
+        googleMap.addMarker(MarkerOptions().position(epfl).title("Marker at EPFL"))
+        googleMap.moveCamera(CameraUpdateFactory.newLatLng(epfl))
+      }
+    }
+  }
```
tsogtb, line 65: Looks neat!

tsogtb (approved):

tsogtb:
Looks so good!

KlaKalma (approved):
Perfect, thanks for taking into account the changes, good comments !

PR #91, opened 2024-04-08
 switch button for the lists and ability to create lists for chosen s…

aymanbakiri:
…ongs and top songs


app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/ProfileViewModel.kt
```diff
@@ -11,7 +11,13 @@ import javax.inject.Inject
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.launch
+import ch.epfl.cs311.wanderwave.model.data.Track
 
+
+
+
+// Define a simple class for a song list
+data class SongList(val name: String, val tracks: MutableList<Track> = mutableListOf())
```
KlaKalma, line 20: Why do you need that ? Unfortunately I think it's gonna be hard to adapt the firebase connection if you use a specific dataclass for a list of tracks. Is it important ? This is a completely open quesiton if you think I should implement the songlist (which you should rename playlist btw), then I'll do it.

gradle/libs.versions.toml
```diff
@@ -1,5 +1,5 @@
 [versions]
-agp = "8.3.0"
+agp = "8.2.2"
```
KlaKalma, line 2: Is is a merge error or it's actually something you need ?

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/ProfileViewModel.kt
KlaKalma: Now that I think about it, the more general choice should be :

- The data class is better if we want "independant" playlists, that live outside of a beacon, that are like use playlists
- Keeping a normal List is better for backend storage and on device storage. and good if we keep playlists for a user or for a beacon.

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/ProfileScreen.kt
```diff
@@ -42,29 +54,109 @@ val INPUT_BOX_NAM_SIZE = 150.dp
...
-  } else {
-
-    Column(modifier = Modifier.fillMaxSize().padding(16.dp).testTag("profileScreen")) {
-      Box(modifier = Modifier.fillMaxWidth()) {
-        VisitCard(Modifier, currentProfile)
-        ProfileSwitch(Modifier.align(Alignment.TopEnd), viewModel)
-        ClickableIcon(Modifier.align(Alignment.BottomEnd), Icons.Filled.Create)
-      }
+fun ProfileScreen(viewModel: ProfileViewModel) {
+    val currentProfileState by viewModel.profile.collectAsState()
```
mbouaissi, line 58: Here, I think what you should do is putting your code from line 85 to the end inside the Box that is line 78. Also, I don't think you should comment the EditableVisitCard here

mbouaissi: And also maybe try to divide your code in multiple function, so it's easier to see what part is doing what

mbouaissi (changes requested):
I think your code is good and easily readable, because it's well commented :+1:. I just think that once he missunderstanding inside ProfileScreen will be solve, we would be able to merge it

gradle/libs.versions.toml
aymanbakiri: I have this error : The project is using an incompatible version (AGP 8.3.0) of the Android Gradle plugin. Latest supported version is AGP 8.2.2
So I was forced to change the agp from 8.3.0 to 8.2.2.
But I think that the problem comes from my android studio version so you can leave agp =8.3.0

mbouaissi (approved):
lgtm

mbouaissi (approved):
lgtm 2

PR #107, opened 2024-04-11
Trigger CI when merging a pull request into main

yzueger:
This will trigger the CI when merging a pull request into the main branch. This would allow us disable branches needing to be up to date before merging with some confidence.


joriba (approved):
Lgtm

PR #111, opened 2024-04-11
Add Anonymous auth

yzueger:
Adds the structure for authentication and a simple implementation using anonymous authentication to Firebase, while we resolve issues with Spotify

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -116,9 +116,12 @@ dependencies {
     kapt(libs.androidx.room.compiler)
     implementation(libs.androidx.room.ktx)
 
-    implementation("com.google.firebase:firebase-core:17.0.0")
-    implementation("com.google.firebase:firebase-database-ktx:20.3.0")
-    implementation("com.google.firebase:firebase-firestore:24.10.0")
+    implementation(platform(libs.firebase.bom))
+    implementation(libs.firebase.core)
+    implementation(libs.firebase.database)
+    implementation(libs.firebase.firestore)
+    implementation(libs.firebase.auth)
+
     implementation("com.google.android.play:core-ktx:1.7.0")
 
     implementation(libs.maps.compose)
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -117,7 +117,6 @@ dependencies {
     implementation(libs.androidx.room.ktx)
 
     implementation(platform(libs.firebase.bom))
-    implementation(libs.firebase.core)
     implementation(libs.firebase.database)
     implementation(libs.firebase.firestore)
     implementation(libs.firebase.auth)
--- a/app/src/main/java/ch/epfl/cs311/wanderwave/model/auth/AuthenticationController.kt
+++ b/app/src/main/java/ch/epfl/cs311/wanderwave/model/auth/AuthenticationController.kt
@@ -1,0 +1,35 @@
+package ch.epfl.cs311.wanderwave.model.auth
+
+import com.google.firebase.auth.FirebaseAuth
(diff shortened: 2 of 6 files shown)


app/src/main/java/ch/epfl/cs311/wanderwave/model/auth/AuthenticationController.kt
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  private data class State(val isSignedIn: Boolean = false)
+
+  data class UserData(
+      val id: String,
+      val email: String?,
+      val displayName: String?,
+      val photoUrl: String?
+  )
```
KlaKalma, line 54: So you need this data class ? Why not use the profile class ? I know it's a bit different but it could be nice to all use the same one. We could add in the email. I can help you with that if you want. Just a question tho

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/LoginScreenViewModel.kt
```diff
@@ -23,16 +31,30 @@ class LoginScreenViewModel @Inject constructor(private val spotifyController: Sp
...
       else -> {
         _uiState.value =
-            UiState(hasResult = true, success = false, message = "User cancelled login")
+            uiState.value.copy(hasResult = true, success = false, message = "User cancelled login")
+      }
+    }
+  }
+
+  private fun authenticate(token: String) {
+    CoroutineScope(Dispatchers.IO).launch {
```
KlaKalma, line 49: Why coroutine scope and not viewmodelscope ?

KlaKalma (changes requested):
Quite unfortunate that we cannot have the classic authentication. Let's see if it works soon or not, code is clean. I just have some little comments. Thanks for that :)

app/build.gradle.kts
```diff
@@ -116,9 +116,11 @@ dependencies {
...
     implementation(libs.androidx.room.ktx)
 
-    implementation("com.google.firebase:firebase-core:17.0.0")
-    implementation("com.google.firebase:firebase-database-ktx:20.3.0")
-    implementation("com.google.firebase:firebase-firestore:24.10.0")
+    implementation(platform(libs.firebase.bom))
+    implementation(libs.firebase.database)
+    implementation(libs.firebase.firestore)
+    implementation(libs.firebase.auth)
+
```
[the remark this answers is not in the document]
yzueger: Sure, I needed to add a dependency there anyway :)

app/src/main/java/ch/epfl/cs311/wanderwave/model/auth/AuthenticationController.kt
yzueger: I do think it's a good idea to separate the parts relevant for authentication from our own data stored about a user. This one corresponds to the authentication data provided by firebase itself (later including Spotify), while the profile corresponds to our own objects in the firestore. But I definitely agree that we should connect them in a nice way. 
For now, the current user profile should have its firebaseUid set to `authenticationController.getUserData().id` to make sure permissions will always work, but with the current structure I'm unsure of where to implement that

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/LoginScreenViewModel.kt
yzueger: I think I once had issues with that because the viewModelScope would stil run on the UI thread, which would not work in some cases. Using the IO scope has worked better so far

PR #139, opened 2024-04-12
Modularize to allow for previews

joriba:
I extracted a BeaconScreen() with no arguments (at the moment, will be filled with information and callbacks that the screen needs), so we can properly test and preview it without duplication.


aymanbakiri (approved):
Thanks for implementing BeaconScreen() with no arguments , it will allow us to preview it without duplication.

PR #142, opened 2024-04-12
aliases example for enhanced productivity

KlaKalma:
Just very simple aliases that I created a while ago and that I thought I would share
They are all very usefull, especially the two new additions today
git cff : automatic commit named ktfmt :)
git rb <branch name> : rebase a branch automatically if possible, else merge it 

How to install is written in the read me

Here is a list of the commands : 
git i : nicer git log
git s : shorter and more efficient git status 
git c : git commit -am 
git undo : reset the last commit


mbouaissi (approved):
lgtm

PR #143, opened 2024-04-13
a user can retrieve the most recently listened title

mbouaissi:
This pull request is only for the user to retrieve his most recently track/album


mbouaissi:
This should only display a song/playlist/album on the profile screen

PR #146, opened 2024-04-16
116 add the play functionality to the spotify controller

IB-12:
Almost ready to merge, I just need to write some tests for it. Otherwise it is fully functional.


mbouaissi (approved):
I think the code looks fine, maybe try to comment it a bit more. About the media player, idk if you only work on the small one, if it's the case, I think I have find a bug, but idk if it's your work. So basically, when you play a song, and switch, the white bar that show how far we are in the song keep going. And also, when we extend the media player I don't see the play and pause buttton. But all these things can be issue for the next sprint

IB-12:
Thank you for your review. I am aware of the issues you've encountered, but unfortunately they're neither part of my previous work or this current issue, so we'll have to solve them in a future sprint.

PR #151, opened 2024-04-18
 addTrackToBeacon + tests

aymanbakiri:


KlaKalma (changes requested):
Why create a new file for this? I don't quite understand why you would need a new file and not just use update beacon from the beaconconnection. This seems a bit overkill, maybe you discussed it without me but I think it would be better to have it directly in the beacon connection file.

PR #153, opened 2024-04-18
 add track to the beacon trough a button in beacon screen

aymanbakiri:


KlaKalma (approved):
Nice thanks, maybe next time be a bit more clear in the name of the pr "add the add button to the beacon screen". Code is nice.

KlaKalma (approved):
Nice thanks lgtm

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/BeaconScreen.kt
```diff
@@ -89,12 +97,12 @@ private fun BeaconScreenPreview() {
 }
 
 @Composable
-private fun BeaconScreen(beacon: Beacon) {
+private fun BeaconScreen(beacon: Beacon, viewModel: BeaconViewModel = hiltViewModel()) {
```
joriba, line 100: My idea with the BeaconScreen here was that it does not need the ViewModel passed, so we can use the @Preview (which does not work with Hilt, I think). 
I don't really know how many parameters it'll be, but if it's not too many, I suggest creating the functions that happen when adding a beacon here, and passing them in as callbacks, instead of passing in the ViewModel. 

If you do find a way to make the preview work with the ViewModel, you could also remove the `beacon` parameter and just get it from the ViewModel

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/BeaconScreen.kt
joriba, line 1: The UI looks really good :)

joriba (changes requested):
I had a comment about passing the ViewModel to the (sub-)BeaconScreen.

joriba:
Forgot to ktfmt ;)

mbouaissi (approved):
Nice code, and thx for the comments, it's easier to review. lgtm :+1:

joriba (dismissed):
Nice, that's better I think :)

mbouaissi (approved):
lgtm

PR #177, opened 2024-04-20
Firebase mocking

### LINKED ISSUE #156: [BUGFIX] Do not test with Firebase in production

KlaKalma:
After making fancy tests with the actual firebase implementation, I deleted all of them and mocked firebase
It implements test for TrackConnection and beaconConnection, testing every function inside.

If you guys have any other idea on how I could do better test tell me, I tried making a more fancy test, with adding and returning complex objects and checking the signatures, but after 3h of trying, I couldn't manage to make it work, check first commit for that.

If you guys change your mind and would like the actual test reinstated, but without the actual clutter on the firebase repo, I think that making functions that clean the repo could be good.

Closes #156


app/src/androidTest/java/ch/epfl/cs311/wanderwave/model/BeaconConnectionTest.kt
```diff
@@ -23,137 +23,70 @@ public class BeaconConnectionTest {
...
-      }
-
-      // Assert that the flow didn't emit anything within the timeout
-      assert(valueEmitted.not()) { "Flow emitted unexpected value" }
-    }
-  }
+  fun testGetItem() = runBlocking {
+    withTimeout(3000) {
+      // Call the function under test
+      val retrievedBeacon = beaconConnection.getItem("testBeacon")
```
yzueger, line 77: This just gets the flow, but we need to also consume it to check that it actually creates a value, eg with a `.first()`. When trying that locally, it doesn't pass yet. This would also help increase coverage.

yzueger (changes requested):
Looking pretty good, just a bit of missing coverage for one of the tests.
Also, shouldn't the ProfileConnectionTest also use a mocked firebase?

KlaKalma:
> Looking pretty good, just a bit of missing coverage for one of the tests. Also, shouldn't the ProfileConnectionTest also use a mocked firebase?

I made the changed but stashed them and didn't put them back thanks for seeing it.

app/src/androidTest/java/ch/epfl/cs311/wanderwave/model/BeaconConnectionTest.kt
KlaKalma: I have tried to make the test better but it's very complicated to make the get item of the beacon work, because it needs to fetch the tracks after it has already fetched the beacon, so it's quite hard to make it work with the mocking we have. It seems like mocking a document like that has some limitations.

IB-12 (approved):
LGTM 👍

KlaKalma:
> LGTM 👍 Edit: Looks like one of the tests is failing, I'm running the CI again just to be sure

It was just that I didn't merge with the main and I still had the problem we had.

PR #179, opened 2024-04-21
Firebase local database implementation

KlaKalma:
This includes : 

- Deleting the local room repositories, the tests that go with it and adapt the view models to use the local repo
- Making the local repository work
- Adapt the tests


app/src/main/java/ch/epfl/cs311/wanderwave/model/remote/TrackConnection.kt
joriba, line 1: We definitely need to declare once and for all what information should be stored on the beacons, and in what way (this affects the documentToItem() function). I will try to create a database schema today for it.

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/MapViewModel.kt
```diff
@@ -35,11 +28,13 @@ constructor(private val repository: BeaconRepositoryImpl, val locationSource: Lo
...
-        _uiState.value = BeaconListUiState(beacons = beacons, loading = false)
-      }
-    }
+    // CoroutineScope(Dispatchers.IO).launch {
+    //   repository.getAll().collect { beacons ->
+    //     _uiState.value = BeaconListUiState(beacons = beacons, loading = false)
+    //   }
+    // }
+
+    // TODO : Implement the repository with the local database
```
joriba, line 37: By 'implement the repository', you mean creating the repository implementation and using it to get the beacons?

joriba (commented):
Hi, I left some comments. I don't think this is completely ready to merge yet, mostly because of the issue that we don't completely know how the database is supposed to look like. I suggest discussing this on tuesday.

Also, I would afterwards consider merging this directly into main... this is work that concerns almost everything.

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/MapViewModel.kt
KlaKalma: Sorry this was a message for the past me, I should delete it, at the time I didn't know how to deal with implementing the firebase local storage, so I just wrote this to remind me to deal with this when I solved the problem. Which I didn't  implement, thanks for pointing this out.

KlaKalma:
Unfortunately, I rebased on the feature branch, I really thought this would fit in better here, and I started working on this branch last week. It's true that probably everyone will need it but I think dealing with the same merge conflicts 2 times is not worth the hassle for now.

app/src/main/java/ch/epfl/cs311/wanderwave/di/DatabaseModule.kt
```diff
@@ -1,25 +0,0 @@
...
-
-  @Provides
-  @Singleton
-  fun provideAppDatabase(@ApplicationContext context: Context): AppDatabase {
-    return Room.databaseBuilder(context.applicationContext, AppDatabase::class.java, "app_database")
-        .createFromAsset("demo_data.db")
-        .fallbackToDestructiveMigration()
-        .build()
-  }
-}
```
yzueger, line 25: I don't think we can completely remove the room database, as we still need to store some local-only data, like authentication tokens.

app/src/main/java/ch/epfl/cs311/wanderwave/di/ConnectionModule.kt
```diff
@@ -0,0 +1,35 @@
...
+  fun provideTrackConnection(@ApplicationContext context: Context): TrackConnection {
+    return TrackConnection()
+  }
+
+  @Provides
+  @Singleton
+  fun provideProfileConnection(@ApplicationContext context: Context): ProfileConnection {
+    return ProfileConnection()
+  }
+}
```
yzueger, line 35: It's a bit nitpicking about the naming, but I would still like to see a sort of repository structure instead of things directly using a "Connection". Then the connections could just implement the repository interface the way the local repositories did before. When things depend only on an abstract interface, it's much easier to add something like a caching layer without the naming becoming very confusing.
we could then have @Provides functions of type XYZRepository still returning the corresponding connection for now.

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/BeaconScreen.kt
```diff
@@ -58,9 +59,11 @@ fun BeaconScreen(
   Column(
       modifier = Modifier.fillMaxSize().padding(16.dp),
       horizontalAlignment = Alignment.CenterHorizontally) {
+        Log.d("BeaconScreen", "UIState: $uiState")
```
mbouaissi, line 62: Haven't we say we don't want any log remaining in the code?

mbouaissi (dismissed):
Good work thanks, I just have small concerns about some part of the code that i have mentioned

app/src/main/java/ch/epfl/cs311/wanderwave/model/repository/BeaconRepositoryImpl.kt
joriba, line 1: This looks more like what I had in mind. What I'm wondering now is:
do we even still need the `BeaconConnection` class? As far as I can see, it's only purpose is to be passed to `BeaconRepositoryImpl` where its functions are called from similarly-named methods. 
Is there any reason to not directly implement that functionality in `BeaconRepositoryImpl`?

The same question applies to the Track and Profile interaction as well, of course.

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/BeaconViewModel.kt
```diff
@@ -4,18 +4,20 @@ import androidx.lifecycle.ViewModel
...
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.launch
 
 @HiltViewModel
-class BeaconViewModel @Inject constructor() : ViewModel() {
-  private val beaconConnection = BeaconConnection()
+class BeaconViewModel @Inject constructor(private val beaconRepository: BeaconRepository) :
+    ViewModel() {
   private val id = "UAn8OUadgrUOKYagf8a2"
```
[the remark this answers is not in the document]
joriba: Yes, this can be removed now, if the mock ViewModel returns something sensible

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/TrackListViewModel.kt
```diff
@@ -29,10 +31,12 @@ constructor(
...
 
   private fun observeTracks() {
-    CoroutineScope(Dispatchers.IO).launch {
-      repository.getAll().collect { tracks ->
-        _uiState.value = UiState(tracks = tracks, loading = false)
-      }
+    viewModelScope.launch {
+      trackRepository.getAll().collect {
+        _uiState.value = _uiState.value.copy(tracks = it, loading = false)
+      } // TODO : update to get only useful tracks
```
joriba, line 37: So this is for the TrackList that just displays ALL tracks right? Then, I think the TODO is not necessary... although the screen itself is kind of obsolete once we have proper beacon- and user-specific track lists.

joriba (changes requested):
Left some comments.
Also, right now, the code doesn't compile, so maybe I missed something that I could only have seen by running the app

app/src/androidTest/java/ch/epfl/cs311/wanderwave/ui/BeaconScreenTest.kt
```diff
@@ -17,15 +24,28 @@ class BeaconScreenTest {
...
+
+    val beaconFlow =
+        flowOf(
+            Beacon(
+                "UAn8OUadgrUOKYagf8a2",
+                Location(46.519653, 6.632273, "Lausanne"),
+                listOf<Track>(Track("Some Track ID", "Track Title", "Artist Name")),
+            ))
+    coEvery { beaconConnection.getItem(any<Beacon>()) } returns beaconFlow
+    coEvery { beaconConnection.getItem(any<String>()) } returns beaconFlow
```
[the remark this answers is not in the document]
KlaKalma: This is because coEvery is for coroutine stuff, and I put it when it has something to do with coroutines, which it absolutely does, dealing with task in the unit test is an anoying mess.
It's not the same, the function is overloaded and I just couldn't put any because it was overloaded so I had to put both.

app/src/main/java/ch/epfl/cs311/wanderwave/model/localDb/AppDatabase.kt
```diff
@@ -1,15 +1,21 @@
...
-  abstract fun trackDao(): TrackDao
 
-  abstract fun profileDao(): ProfileDao
 
-  abstract fun beaconDao(): BeaconDao
+@Database(entities = [PlaceHolderEntity::class], version = 1)
+abstract class AppDatabase : RoomDatabase() {
 }
+
+// place holder entity
```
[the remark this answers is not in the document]
KlaKalma: THe app database needs something in it, else it doesn't compile. The team told me to leave the database so I just put a place holder, it's just temporary until you store the tokens etc...

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/TrackListViewModel.kt
KlaKalma: I mean, it was more todo : make a function that can fetch tracks that are only from a certain beacon etc.... It's true that it will be obsolete soon

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/BeaconViewModel.kt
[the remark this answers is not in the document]
KlaKalma: I didn't add it

app/src/main/java/ch/epfl/cs311/wanderwave/model/repository/BeaconRepositoryImpl.kt
KlaKalma: THe BeaconConnection have the firebase connection as a superclass, I would really like to keep it like that because it's very easy to implement new connections to firebase like that, so how would you deal with that ? Make something like
BeaconConnection : FirebaseConnection : BeaconRepositoryImpl : BeaconRepository ?
Seems a bit fancy no ? So the best way I had was to do like that.

KlaKalma: Done

joriba (approved):
lgtm

PR #184, opened 2024-04-22
[Feature] Spotify auth

### LINKED ISSUE #69: Connect spotify authentication to firebase

yzueger:
Closes #69


app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/LoginScreenViewModel.kt
```diff
@@ -23,16 +31,30 @@ class LoginScreenViewModel @Inject constructor(private val spotifyController: Sp
...
       else -> {
         _uiState.value =
-            UiState(hasResult = true, success = false, message = "User cancelled login")
+            uiState.value.copy(hasResult = true, success = false, message = "User cancelled login")
+      }
+    }
+  }
+
+  private fun authenticate(token: String) {
+    CoroutineScope(Dispatchers.IO).launch {
```
KlaKalma, line 49: I know I'm bothering many people with this but I asked chatGPT multiple times about it and it really seems better to have the viewmodel scope instead of normal coroutine scope: Here is chatGPT answer
Error handling: viewModelScope has a default SupervisorJob installed. This means that if one child job fails, other children are not affected and continue their execution.  
Memory leaks: Since viewModelScope is tied to the ViewModel lifecycle, it helps prevent memory leaks by cancelling all running coroutines when the ViewModel is cleared.

KlaKalma (commented):
Very nice thanks, seems like a lot of work ! Code is very clean

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/LoginScreenViewModel.kt
yzueger: #208

joriba (approved):
lgtm

PR #185, opened 2024-04-22
[Spotify auth] Get token from Firebase Functions

yzueger:
This PR changes the Firebase authentication from anonymous to using Spotify via the Firebase Functions.


KlaKalma (approved):
Very nice :)

PR #187, opened 2024-04-23
Create the data class ProfileTrackAssociation

mbouaissi:
This is a very small PR to just create the data class that associate a user with a track


mbouaissi:
The thing is that in a beacon a user can maybe only add one song, so we don't need to have a list of song in my opinion

IB-12 (approved):
lgtm

PR #189, opened 2024-04-23
[Bugfix] Lower accompanist version to 0.34.0

yzueger:
On my device, attempting to display the TrackList screen causes a crash, including for tests. This issue is also described here: https://github.com/google/accompanist/issues/1752
The described solution is lowering the version of accompanist to 0.34.0, which is also working in my case.


mbouaissi:
lgtm

yzueger:
@mbouaissi could you mark the review as approved :)

KlaKalma (approved):
lgtm

PR #190, opened 2024-04-23
[Spotify auth] Complete authentication controller with refresh tokens

yzueger:
This PR completes the authentication controller functionalities, including storing and using refresh tokens for extended login sessions. It does not include linking the users profile to the correct Firebase id and similar connections, which are probably best implemented in a separate feature, once the local firestore repositories are also merged.
After this PR the Spotify auth feature can be merged into main


aymanbakiri (approved):
LGMT, great code

PR #191, opened 2024-04-23
134 add the loop functionality to the spotify controller

IB-12:


app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/TrackListViewModel.kt
```diff
@@ -152,13 +160,18 @@ constructor(
     skip(-1)
   }
 
+  fun toggleLoop() {
+    _uiState.value = _uiState.value.copy(isLooping = !_uiState.value.isLooping)
```
KlaKalma, line 164: I think it would be better if you put some kind of way to have it turn to True or false instead of only switching from one.
You could overload the function or put a default argument `fun toggleLoop(val isLooping = !_uiState.value.isLooping) { ...`

KlaKalma (commented):
Nice

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/TrackListViewModel.kt
IB-12: Will do, thanks for the suggestion

aymanbakiri (approved):
FGMT

PR #194, opened 2024-04-24
go to the beacon screen when clicking on a beacon

joriba:
Right now, there are some invalid beacons still showing on the map (they have a wrong schema, so clicking them goes to a blank beacon screen).

### DIFF
--- a/app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/map/BeaconMapMarker.kt
+++ b/app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/map/BeaconMapMarker.kt
@@ -8,6 +8,6 @@ import com.google.maps.android.compose.MarkerState
 
 @Composable
 @GoogleMapComposable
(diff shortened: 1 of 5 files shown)


app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/map/BeaconMapMarker.kt
```diff
@@ -8,6 +8,18 @@ import com.google.maps.android.compose.MarkerState
...
+    title: String? = null,
+    snippet: String? = null,
+    onClick: () -> Unit = {}
+) {
+  Marker(
+      state = MarkerState(position = position),
+      title = title,
+      snippet = snippet,
+      onClick = {
+        onClick()
```
mbouaissi, line 22: I dont really understand why do we have true here?

mbouaissi (approved):
Apart from the small comment i have left, but yeah it works very well :+1:

PR #196, opened 2024-04-24
remove the "sample beacon" to show a loading animation

joriba:
The loading animation was gone because the viewmodel was initialized with a dummy beacon. I removed it in favor of `null` and setting `isLoading` to `true`, so now the animation is back.


IB-12 (approved):
lgtm

PR #197, opened 2024-04-24
[Feature] Create a beacon screen and the framework behind

KlaKalma:
This contains : 
- The actual beacon screen
- The new beacon and firebase local implementation
- The new beacon screen view model
- The different tests for the beacon screen


joriba (approved):
Everything already has been reviewed at some point, so should be fine, right?

PR #224, opened 2024-04-27
Feature/place beacons/place randomly

### LINKED ISSUE #205: Have the places with no beacons automatically create them

mbouaissi:
This PR add the functionnality to add beacons near a position if there is not enough beacons. It uses randomness to try to make the beacon distribution as fair as possible. 

Closes #205

### DIFF
--- a/app/src/androidTest/java/ch/epfl/cs311/wanderwave/utils/BeaconPlacerTest.kt
+++ b/app/src/androidTest/java/ch/epfl/cs311/wanderwave/utils/BeaconPlacerTest.kt
@@ -1,6 +1,7 @@
 package ch.epfl.cs311.wanderwave.utils
 
 import android.util.Log
+import androidx.test.ext.junit.runners.AndroidJUnit4
 import ch.epfl.cs311.wanderwave.model.data.Beacon
 import ch.epfl.cs311.wanderwave.model.data.Location
 import ch.epfl.cs311.wanderwave.model.utils.computeDistanceBetweenBeacons
@@ -14,9 +15,11 @@ import com.kaspersky.kaspresso.testcases.api.testcase.TestCase
 import junit.framework.TestCase.assertEquals
 import junit.framework.TestCase.assertTrue
 import org.junit.Test
+import org.junit.runner.RunWith
 
 private const val BEACON_COUNT = 20
 
+@RunWith(AndroidJUnit4::class)
 class BeaconPlacerTest : TestCase() {
 
   @Test
--- a/app/src/androidTest/java/ch/epfl/cs311/wanderwave/utils/BeaconPlacerTest.kt
+++ b/app/src/androidTest/java/ch/epfl/cs311/wanderwave/utils/BeaconPlacerTest.kt
@@ -1,0 +1,128 @@
+package ch.epfl.cs311.wanderwave.utils
+
+import android.util.Log
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import ch.epfl.cs311.wanderwave.model.data.Beacon
+import ch.epfl.cs311.wanderwave.model.data.Location
+import ch.epfl.cs311.wanderwave.ui.components.utils.computeDistanceBetweenBeacons
+import ch.epfl.cs311.wanderwave.ui.components.utils.countNearbyBeacons
+import ch.epfl.cs311.wanderwave.ui.components.utils.findRandomBeacon
+import ch.epfl.cs311.wanderwave.ui.components.utils.haversine
+import ch.epfl.cs311.wanderwave.ui.components.utils.placeBeaconsRandomly
+import ch.epfl.cs311.wanderwave.ui.components.utils.randomLatLongFromPosition
+import com.google.android.gms.maps.model.LatLng
+import com.kaspersky.kaspresso.testcases.api.testcase.TestCase
+import junit.framework.TestCase.assertEquals
+import junit.framework.TestCase.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+
+private const val BEACON_COUNT = 20
+
+@RunWith(AndroidJUnit4::class)
+class BeaconPlacerTest : TestCase() {
+
+    @Test
+    fun placeBeaconRandomly_addsBeaconWhenCountIsLessThanThreshold() {
+        val location = LatLng(46.519962, 6.633597)
+        val beacons = listOf<Beacon>()
+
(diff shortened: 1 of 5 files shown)


app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/utils/BeaconPlacer.kt
```diff
@@ -0,0 +1,98 @@
...
+import kotlin.math.*
+import kotlin.random.Random
+
+private const val EARTH_RADIUS_M = 6371000.0
+private const val BEACON_RADIUS = 1000.0
+private const val BEACON_COUNT = 20
+private const val NUMBER_ITERATION = 5
+
+fun placeBeaconsRandomly(beacons: List<Beacon>, location: LatLng): List<Beacon> {
+  var finalBeacons = mutableListOf<Beacon>()
```
IB-12, line 15: If you could add some comments it would be quite helpful I think, it'd make it easier to understand what happens

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/utils/BeaconPlacer.kt
```diff
@@ -0,0 +1,98 @@
...
+    }
+  }
+  return distance
+}
+
+fun findRandomBeacon(location: LatLng, newBeacons: MutableList<Beacon>, it: Int) {
+  val randomLocation = randomLatLongFromPosition(location, BEACON_RADIUS)
+  val newBeacon =
+      Beacon(
+          id = "beacon${newBeacons.size + it}", // TODO: modify the ID!!!!!!!!
```
IB-12, line 49: Can't you just use Firebase to generate a new ID ?

IB-12 (dismissed):
You've done a good job, and your tests seem fairly exhaustive 👍 In my opinion this code should also handle adding the new beacons to Firebase, which would in turn allow you to easily generate a unique ID for them, but let me know what you think.

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/utils/BeaconPlacer.kt
KlaKalma, line 49: Put an empty id and add it to firebase, I can create you a function that will add the beacon to firebase and return the new ID. Tell me if you need it, I'll push it on your branch if you need :)

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/utils/BeaconPlacer.kt
```diff
@@ -0,0 +1,98 @@
+package ch.epfl.cs311.wanderwave.ui.components.utils
+
+import ch.epfl.cs311.wanderwave.model.data.Beacon
+import ch.epfl.cs311.wanderwave.model.data.Location
+import com.google.android.gms.maps.model.LatLng
+import kotlin.math.*
+import kotlin.random.Random
+
+private const val EARTH_RADIUS_M = 6371000.0
```
KlaKalma, line 9: Should we put this in the ressources ? I think it's fine as it's only related to this file, but we may have to do it, if we need it soon in another file

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/utils/BeaconPlacer.kt
```diff
@@ -0,0 +1,98 @@
+package ch.epfl.cs311.wanderwave.ui.components.utils
```
KlaKalma, line 1: I don't think this should go there, I agree it's an utils file, but not at all for ui only. This seems much more like a background that the user may not even be aware or a little aware.

app/src/androidTest/java/ch/epfl/cs311/wanderwave/utils/BeaconPlacerTest.kt
```diff
@@ -0,0 +1,158 @@
...
+import com.google.android.gms.maps.model.LatLng
+import com.kaspersky.kaspresso.testcases.api.testcase.TestCase
+import junit.framework.TestCase.assertEquals
+import junit.framework.TestCase.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+
+private const val BEACON_COUNT = 20
+
+@RunWith(AndroidJUnit4::class)
```
KlaKalma, line 22: Do you need this ? It seems like a unit test so why running it with emulator like this ?

KlaKalma (commented):
Simple and clean algorithm, thanks for that, just some little comments. Just tell me if you need the function I'll push it on the branch, or make a new one if you want to be finished with this one !

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/utils/BeaconPlacer.kt
KlaKalma: Yep would be much better please !

mbouaissi: Yes please, it will be much appreciated

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/utils/BeaconPlacer.kt
mbouaissi: True, if we use it somewhere else, we should add a file for the constants

mbouaissi:
Thanks for your review guys :+1:.  I'm sorry, when i pushed, i totally forgot to comment the code. Also, when i will have the firebase function for the id, i could integrate it to these functions

app/src/androidTest/java/ch/epfl/cs311/wanderwave/utils/BeaconPlacerTest.kt
mbouaissi: so actually for some really really strange reasons, when i remove it, the test from the trackListScreen where not passing (the two class are totally unrelated), so for now i think we should keep it

aymanbakiri (approved):
Your code is great and well-commented.
Using randomness to make the beacon distribution as fair as possible is perfect.
Also, thanks for moving the file from the UI folder since it is more related to the backend.
The only issue was that you were not using Firebase to create a new beacon ID, but now you have corrected this.
I think it is ready to merge.
Great job.

mbouaissi:
Thanks for your review guys

PR #225, opened 2024-04-28
Added beacon generation using Google's Places

IB-12:
This allows to create beacons around the user using Google's Places. We need to define:

- [ ] What kind of places we want to use to create new beacons
- [ ] The minimal distance between any two beacons


app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/utils/BeaconGenerator.kt
KlaKalma, line 1: Same as before I think it's not a good place to put something that's not really realted to ui in the ui folder, especially as those feature are very much related to the backend

aymanbakiri (changes requested):
Your code is good ,very clear and well commented. Also you have done fairly exhaustive tests.
Now we can create beacons around the user using Google's Places.
I think that small change that you need to do is to move the BeaconGenerator from the ui folder and put it in the backEnd folder instead.
Good job , thanks

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/utils/BeaconGenerator.kt
IB-12: You're right, I corrected it. (Also applies to @aymanbakiri 's comment)

app/src/main/java/ch/epfl/cs311/wanderwave/model/data/Location.kt
mbouaissi, line 1: Good idea of putting the function of distanceBetween inside the location instead. When we are merging we can change this.

app/src/main/java/ch/epfl/cs311/wanderwave/model/utils/BeaconGenerator.kt
```diff
@@ -0,0 +1,133 @@
...
+import com.google.android.libraries.places.api.model.Place
+import com.google.android.libraries.places.api.net.FindCurrentPlaceRequest
+
+// TODO: What would be a good value for these ?
+private const val MIN_REVIEWS = 100
+private const val MIN_BEACON_DISTANCE = 100.0
+
+// TODO: Store as a resource
+// TODO: Which types do we want ?
+val types: List<String> =
```
mbouaissi, line 22: Good idea, we should discuss in group if some of them are irrelevant or if we should add more (if possible)

app/src/main/java/ch/epfl/cs311/wanderwave/model/utils/BeaconGenerator.kt
```diff
@@ -0,0 +1,133 @@
...
+  // Get the nearby points of interest
+  val nearbyPOIs = getNearbyPOIs(context, location, radius)
+
+  // Place new beacons at the nearby points of interest
+  for (poi in nearbyPOIs) {
+    // Check if POI is far enough from existing beacons
+    if (nearbyBeacons.any() { beacon ->
+      beacon.location.distanceBetween(poi) > MIN_BEACON_DISTANCE
+    }) {
+      newBeacons.add(Beacon("", poi))
```
mbouaissi, line 70: For now it's fine, but when @KlaKalma will have prepare the corresponding method, we should look back at this for the id

app/src/main/java/ch/epfl/cs311/wanderwave/model/utils/BeaconGenerator.kt
```diff
@@ -0,0 +1,133 @@
...
+                } // Check if the place is of a certain type
+            ) {
+              nearbyPOIs.add(placeLoc)
+              Log.d("PlacesApi", "Place found: ${place.name} at ${place.latLng}")
+            }
+          }
+        }
+        .addOnFailureListener { exception ->
+          if (exception is ApiException) {
+            Log.d("PlacesApi", "Place not found: ${exception.statusCode}")
```
mbouaissi, line 127: In this case, Log.e is better i think for the errors

mbouaissi: but its just a detail lol

mbouaissi (approved):
Very nice code, it's well commented and well structured. The comments I have made are more suggestions, so it's up to you!

app/src/main/java/ch/epfl/cs311/wanderwave/model/utils/BeaconGenerator.kt
IB-12: You're right, I'll edit it

app/src/main/java/ch/epfl/cs311/wanderwave/model/utils/BeaconGenerator.kt
```diff
@@ -0,0 +1,149 @@
...
+            }
+          }
+        }
+        .addOnFailureListener { exception ->
+          if (exception is ApiException) {
+            Log.e("PlacesApi", "Place not found: ${exception.statusCode}")
+            Log.e("PlacesApi", "Place not found: ${exception.message}")
+            Log.e("PlacesApi", "Place not found: ${exception.localizedMessage}")
+            Log.e("PlacesApi", "Place not found: ${exception.cause}")
+          }
```
tsogtb, line 145: maybe just log that it wasn't an ApiException

tsogtb (approved):
Overall, looks neat nicely done.

aymanbakiri (commented):
forget to approve

aymanbakiri (approved):
As I said in my previous review, your code is good and very clear.
Now we can create beacons around the user using Google's Places.
Thanks for moving the BeaconGenerator from the ui folder to the backEnd folder instead.
I think it is ready to merge.
Good job , thanks

PR #226, opened 2024-04-29
201 consistent track list

joriba:
Add a separate TrackList component that manages the look of the track list. 
The component is now used from the Beacon Screen, the Tracklist screen and can also be used in the Profile screen (WIP). 
This makes it easier to add new features to a track list, like displaying album covers or users that created the tracks (on the beacons, for example, although we might want to use overloading for that instead).


app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/BeaconScreen.kt
```diff
@@ -143,83 +123,16 @@ fun BeaconInformation(location: Location) {
...
-                text = track.artist,
-                color = MaterialTheme.colorScheme.onSurfaceVariant,
-                style = MaterialTheme.typography.bodyMedium,
-            )
+  TrackList(
+      beacon.tracks,
+      title = stringResource(R.string.beaconTracksTitle),
+      onAddTrack = {
+        addTrackToBeacon(beacon.id, it) { success ->
+          if (success) {
```
mbouaissi, line 131: except debug purposes, is there a reason why would we want to check the success here?

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/profile/TrackActions.kt
```diff
@@ -129,28 +130,19 @@ fun AddTrackDialog(
...
-        ?.let { songList ->
-          if (songList.tracks.isNotEmpty()) {
-            Text("CHOSEN SONGS")
-            TracksList(songList.tracks)
-          } else {
-            Text("The CHOSEN SONGS List is empty")
-          }
-        }
-  }
+fun SongsListDisplay(
```
mbouaissi, line 133: nice :+1:

mbouaissi (approved):
Nice work :+1:

mbouaissi (approved):
lgtm

PR #230, opened 2024-04-30
Add the beacon Icon and put it on the map

KlaKalma:
This Aims at adding the beacon ICON to the map


IB-12 (approved):
Nice work on the icon👍

PR #231, opened 2024-04-30
Feature/profile integration/profile private public mode

aymanbakiri:
When you select a beacon on the map and attempt to view the associated profile, access is granted only if the profile is set to public. If the profile is private, a message will be displayed, informing you that the profile is private and access to the profile information is restricted.


mbouaissi (changes requested):
Thanks for adding this mode. I just think that is better to have an actual pop up rather than a Log.d, because the user is not suppose to see the log.d, only us when debugging can see it

aymanbakiri (commented):
I removed the log.d and replace it by a pop up

mbouaissi (dismissed):
thanks for the change, lgtm :+1:

mbouaissi (approved):
lgtm 2

PR #232, opened 2024-04-30
 add the beacon to firebase

### LINKED ISSUE #203: adapt the beaconConnection after we have add the profile-track mapping

aymanbakiri:
Through this function , we are now able to add a beacon to firebase.
close #203


joriba (changes requested):
When adding the tracks to the new beacon, the tracks should also contain information on who created them (this should also be done in `addTrackToBeacon`, so maybe you can even just call that after the beacon has been created). I think the way it is now, the created beacon will be invalid and crash the app, unfortunately.

Otherwise, good job.

KlaKalma (changes requested):
Why do you don't change the add function? It already exists, if you want to change the behaviour of the add function then you should change addItem, not create a new function, I understand now we need to add the references of the track to the beacon but there is also a function for that (add track to beacon) So the function you implemented is rundondant I think.

PR #234, opened 2024-05-01
Unify usage of coroutine scope across project

### LINKED ISSUE #208: Use same coroutine scope everywhere

yzueger:
Closes #208 
The refactors the coroutine scope usage so that basically everything is callable from a viewmodel coroutine scope. Only in the actual model implementations of repositories that directly talk to a database/webserver or something we switch to the IO sope using `withContext(Dispatchers.IO) {}`


app/src/main/java/ch/epfl/cs311/wanderwave/model/auth/AuthenticationController.kt
```diff
@@ -66,7 +66,9 @@ constructor(
             .post("code=$authenticationCode".toRequestBody())
             .build()
 
-    val responseJson = httpClient.newCall(request).execute().body?.string() ?: return false
+    val responseJson =
+        withContext(Dispatchers.IO) { httpClient.newCall(request).execute().body?.string() }
```
tsogtb, line 70: I think we can inject dispatchers according to this
https://developer.android.com/kotlin/coroutines/coroutines-best-practices#inject-dispatchers

tsogtb (dismissed):

IB-12 (approved):
Sonarcloud doesn't like us setting the dispatchers ourself, but I'm not sure there really is a way around that. LGTM, make sure to add the needed coverage and I'll approve it again 👍

Edit: Hadn't seen Tsogt's post

aymanbakiri (commented):
Thanks for updating the coroutine scope usage in our project. Now, everything can be called from a viewModel coroutine scope. You have also corrected the issue of manually setting the dispatchers by injecting dispatchers instead. I think you only need to increase the coverage and it will be ready to merge. Thanks for your good work.

KlaKalma (approved):
Seems perfect, thanks for that :)

PR #239, opened 2024-05-02
End to end/media player

IB-12:
I am able to test what's on the AppScreen (namely the scaffold and bottom bar) and the various screens separately, but not both things in the same test, which basically means I am stuck.


app/src/androidTest/java/ch/epfl/cs311/wanderwave/endToEnd/LoginAndUseMediaControllerEndToEndTest.kt
```diff
@@ -0,0 +1,277 @@
...
+
+      loopButton.assertIsDisplayed()
+      loopButton.performClick()
+      loopButton.performClick()
+    }
+  }
+}
+
+/*
+package ch.epfl.cs311.wanderwave.endToEnd
```
mbouaissi, line 130: Can you please remove the commented stuff, once it's done, lgtm

mbouaissi (changes requested):
once no comment, lgtm

PR #249, opened 2024-05-02
233 playable tracks

joriba:
This makes all displayed Track Lists able to play songs from it. 
To achieve this, a lot of the code in SpotifyController was refactored; namely, the logic for playing, skipping etc was implemented in TrackListViewModel before, which is bad, since then, multiple Screens need this viewModel.
Now, that logic is directly implemented on the Spotify controller, which can be passed to the different viewmodels, that can then expose functions to their respective screens. In particular, the MiniPlayer/ExclusivePlayer now have their own ViewModel, PlayerViewModel, which exposes most of the start and skip etc. functionality.
The TrackListViewModel, BeaconListViewModel and ProfileViewModel only expose functions to select a song, which selects the song in the spotifyController.


app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/PlayerViewModel.kt
```diff
@@ -0,0 +1,86 @@
...
+
+  fun skipForward() {
+    viewModelScope.launch { spotifyController.skip(1) }
+  }
+
+  fun skipBackward() {
+    viewModelScope.launch { spotifyController.skip(-1) }
+  }
+
+  fun toggleShuffle() {
```
mbouaissi, line 66: maybe remove the commented code?

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/SelectSongScreen.kt
```diff
@@ -81,3 +89,24 @@ fun SelectSongScreen(navActions: NavigationActions, viewModel: SpotifySongsActio
             }
       }
 }
+
+// TODO :Merge with @Joriba stuff if possible
```
mbouaissi, line 93: For this, idk if this is possible to maybe merge with your previous work? (if not it's fine, just remove the TODO)

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/tracklist/TrackListItem.kt
```diff
@@ -63,3 +100,93 @@ fun TrackListItem(track: Track, selected: Boolean, onClick: () -> Unit) {
         }
       }
 }
+
+/**
+ * Same as [TrackListItem] but with a profile picture on the right side of the track information.
+ * Takes a [ProfileTrackAssociation] instead of a [Track].to display profile information, as well as
+ * a [NavigationActions] to navigate to the profile view screen.
+ */
+@Composable
```
mbouaissi, line 109: Nice

app/src/main/java/ch/epfl/cs311/wanderwave/model/spotify/SpotifyController.kt
```diff
@@ -97,34 +99,95 @@ class SpotifyController(private val context: Context) {
...
+      callbackFlow {
+        val callbackResult =
+            appRemote
+                ?.playerApi
+                ?.subscribeToPlayerState()
+                ?.setEventCallback {
+                  trySend(it)
+                  startPlaybackTimer(it.track.duration)
+                }
+                ?.setErrorCallback { Log.d("SpotifyController", "Error in player state flow") }
```
mbouaissi, line 185: (it's a detail, but i think Log.e is better for this kind of stuff)

mbouaissi (changes requested):
Thank you very much for the code, it's nice to see the track being playable now! I just left a few comments, but once they are addressed, i think it should be good to merge. Thanks for your code

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/PlayerViewModel.kt
joriba: Well, rather than remove it, I probably need to implement it :)
Done, the Shuffling logic is now done in the SpotifyController.

app/src/main/java/ch/epfl/cs311/wanderwave/model/spotify/SpotifyController.kt
joriba: Fixed in the next commit

mbouaissi (changes requested):
Ok thank you very much for the code, you are nearly there ;) But there is still a few issues:
1. Why can't I see the name of the songs anymore?
Screenshot_20240517-133423_wanderwave
2. There is an issue related to the profile screen and tracklist screen. When you add a song you can't play it. I think this might be due to the fact that when we add a song in these place, we don't add them on firebase, so maybe by adding these functions we can make these songs playable(if you need help for this, i can help you)
Otherwise the code looks fine for me

IB-12:
Just wanted to confirm that everything works flawlessly now, I can access the SelectSongScreen from the profile, browse the playlists, add a song and play it right away.

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/player/ExclusivePlayer.kt
```diff
@@ -230,16 +227,16 @@ fun PlayPauseButton(viewModel: TrackListViewModel, uiState: TrackListViewModel.U
...
 
 @Composable
-fun RepeatButton(viewModel: TrackListViewModel, uiState: TrackListViewModel.UiState) {
-  IconButton(onClick = { viewModel.toggleLoop() }, modifier = Modifier.testTag("toggleRepeat")) {
-    when (uiState.loopMode) {
-      LoopMode.NONE ->
+fun RepeatButton(viewModel: PlayerViewModel, uiState: PlayerViewModel.UiState) {
+  IconButton(onClick = { viewModel.toggleRepeat() }, modifier = Modifier.testTag("toggleRepeat")) {
+    when (uiState.repeatMode) {
+      false ->
```
IB-12, line 233: Was there any particular reason for removing the ability to loop over one single track ? It's a feature that is available in most (if not all) media players so ideally we should keep it.

IB-12 (changes requested):
Overall you did a very good job with this massive refactoring. The only two issues I have noticed are the following:
- The media player is not visible in the BeaconScreen, which isn't great if we want to play songs from there
- It looks like you've removed the "loop once" functionality and I don't really understand why
Otherwise everything looks good 👍

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/player/ExclusivePlayer.kt
IB-12: I can do it if it's too much of a hassle, let me know

joriba: Yeah I did, but now the tests are acting weird. There's not really any useful output on the CI, but when I run the tests locally, a seemingly random one fails each time, also with minimal output... Have you experienced this before?
I just tried running 'All tests' locally, and the first time, makeApiRequest_ShouldReturnExpectedResult_OnSuccessfulApiCall() failed, the second time it was the e2e-test (timeout I think) and the third time spotifyLoginRunsIntent(). Very strange

joriba: Edit: I have the same problem on the main branch as well, locally, so maybe it's unrelated.
Are you able to run the tests on your device consistently? Does anything fail there?

IB-12 (approved):
Just went through your changes and everything looks good to me, well done 👍

mbouaissi (approved):
lgtm, thx

joriba:
Issue for additional coverage: #341 
(Not _that_ urgent though)

PR #253, opened 2024-05-03
Feature/place beacons/main

mbouaissi:
This PR just add the  beacon placement to the main


app/src/main/java/ch/epfl/cs311/wanderwave/model/utils/BeaconPlacer.kt
```diff
@@ -0,0 +1,164 @@
...
+ *
+ * @param position1 the first location
+ * @param position2 the second location
+ * @return the haversine distance between the two locations in meters
+ * @see <a href="https://en.wikipedia.org/wiki/Haversine_formula">Haversine formula</a>
+ * @author Menzo Bouaissi
+ * @since 2.0
+ * @last update 2.0
+ */
+fun haversine(position1: LatLng, position2: LatLng): Double {
```
IB-12, line 95: I don't think we need this function anymore since there is one in `Location.kt`

IB-12 (changes requested):
LGTM, there's just a redundant function that needs to be removed

IB-12 (approved):
Thanks for the change, lgtm

PR #255, opened 2024-05-04
Firebase connection refactoring and songs can be stored in the profile

KlaKalma:
These change profile, beacon and track connection, and the respective repositories.

The changes are the following : 
- Refactoring of add and get item from firebase connection to allow for more adaptability and less repetitions of code
- Added fetch track function in track connection to fetch the references in other objects
- Added the get item in the profile that can retrieve the different songs stored as references in the profile
- Added the add item in the profile that adds a song when it's not in the tracks list


app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/LoginScreen.kt
```diff
@@ -56,10 +63,38 @@ fun LoginScreen(
     LoginAppLogo(modifier = Modifier.weight(1f))
     WelcomeTitle(modifier = Modifier.weight(4f))
     SignInButton(modifier = Modifier.weight(1f)) {
-      val intent =
-          AuthorizationClient.createLoginActivityIntent(
-              context.getActivity(), viewModel.getAuthorizationRequest())
-      launcher.launch(intent)
+      // val intent =
```
mbouaissi, line 66: is it normal that this is commented?

app/src/main/java/ch/epfl/cs311/wanderwave/model/remote/BeaconConnection.kt
```diff
@@ -60,81 +60,51 @@ class BeaconConnection(
...
+  override fun getItem(
+      itemId: String,
+      onSuccess: (DocumentSnapshot, MutableStateFlow<Beacon?>) -> Unit
+  ): Flow<Beacon> {
+    val onSuccessWrapper: (DocumentSnapshot, MutableStateFlow<Beacon?>) -> Unit =
+        { document, dataFlow ->
+          val beacon =
+              dataFlow.value
+                  ?: Beacon.from(document)
+                  ?: Beacon(
```
mbouaissi, line 75: Why are you doing this? Are you placing a beacon at the position (0,0) if we fail to retrieve the beacon?

mbouaissi (changes requested):
Thanks for refactoring the code of the firebase functions. Apart from the commented code and the way you handle the beacons, the rest seems fine in my opinion, thanks :+1:

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/LoginScreen.kt
KlaKalma: This is just for testing purposes, I will get rid of it once the or is ready and I did the testing

app/src/main/java/ch/epfl/cs311/wanderwave/model/remote/BeaconConnection.kt
KlaKalma: It's true, I'll return null in this case

KlaKalma: Solved

app/src/main/java/ch/epfl/cs311/wanderwave/model/remote/FirebaseConnection.kt
```diff
@@ -87,29 +87,38 @@ abstract class FirebaseConnection<T, U> {
...
         .addOnSuccessListener { document ->
           if (document != null && document.data != null) {
-            documentToItem(document)?.let { dataFlow.value = it }
+            documentToItem(document)?.let {
+              dataFlow.value = it
+              onSuccess(document, dataFlow)
+            }
           }
         }
         .addOnFailureListener { e -> Log.e("Firestore", "Error getting document: ", e) }
```
yzueger, line 124: I don't understand why there is an onSuccess parameter, you should just be able to collect the value from the flow right?

Also the error case should probably be propagated to the flow, so that it's not possible to deadlock waiting for a failed item forever. Might be easier with the `flow { }` syntax.

app/src/main/java/ch/epfl/cs311/wanderwave/model/remote/ProfileConnection.kt
```diff
@@ -41,14 +47,26 @@ class ProfileConnection(private val database: FirebaseFirestore? = null) :
...
+    trackConnection.addItemsIfNotExist(item.chosenSongs)
+  }
+
+  override fun addItemWithId(item: Profile) {
+    super.addItemWithId(item)
+    trackConnection.addItemsIfNotExist(item.topSongs)
+    trackConnection.addItemsIfNotExist(item.chosenSongs)
+  }
+
+  fun addProfilesIfNotExist(profiles: List<Profile?>) {
```
yzueger, line 67: I'm not sure in what situation a user would even have permission to add multiple profiles to the database, what the usecase here? But not directly part of this PR.

app/src/main/java/ch/epfl/cs311/wanderwave/model/repository/FirebaseRepository.kt
```diff
@@ -21,7 +22,7 @@ interface FirebaseRepository<T> {
 
   fun deleteItem(itemId: String)
 
-  fun getItem(item: T): Flow<T>
-
   fun getItem(itemId: String): Flow<T>
+
+  fun getItem(itemId: String, onSuccess: (DocumentSnapshot, MutableStateFlow<T?>) -> Unit): Flow<T>
```
yzueger, line 27: As written before somewhere, why the onSuccess?

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/LoginScreen.kt
```diff
@@ -60,6 +60,35 @@ fun LoginScreen(
...
+      //     BeaconConnection(trackConnection = trackConnection, profileConnection =
+      // profileConnection)
+
+      // profileConnection.addItemWithId(profile)
+      //
+      // GlobalScope.launch {
+      //   profileConnection.getItem(profile.firebaseUid).collect {
+      //     Log.d("Debug", "collected Profile ${it}")
+      //   }
+      // }
```
yzueger, line 91: Commented out code

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/ProfileViewModel.kt
```diff
@@ -108,15 +111,18 @@ constructor(
    * @since 2.0
    * @last update 2.0
    */
-  fun retrieveTracks() {
-    viewModelScope.launch {
+  fun retrieveTracksFromSpotify(scope: CoroutineScope = CoroutineScope(Dispatchers.IO)) {
```
yzueger, line 114: I thought we decided to use the `viewModelScope` in the ViewModels, and only use `Dispatchers.IO` when necessary in the repository implementations. 
If we also want to inject scopes in the viewModel, I suggest doing so in the constructor, and setting the default value to `viewModelScope`

KlaKalma: I didn't do it but I can change it

app/src/main/java/ch/epfl/cs311/wanderwave/model/repository/FirebaseRepository.kt
KlaKalma: This is because of the refactoring, in the firebase repository, I have only the base function and I have the beacon and profile connection that sometimes need to do something after the success for example fetch all the references for tracks.

app/src/main/java/ch/epfl/cs311/wanderwave/model/remote/ProfileConnection.kt
KlaKalma: This was just to make sure that the trackAndProfile have some data but it's true that now that the profile can be empty in the track and profile assotiation, we can delete this line ! For another issue next week

app/src/main/java/ch/epfl/cs311/wanderwave/model/remote/FirebaseConnection.kt
KlaKalma: It's not for the user, it's more the other firebase function. This is the best solution I found, to do this. I can explain it to you more, but if you want to understand you can check the implementations of getitem for the beacon and profile connections.

app/src/main/java/ch/epfl/cs311/wanderwave/model/data/Track.kt
```diff
@@ -23,6 +24,7 @@ data class Track(
 
   companion object {
     fun from(document: DocumentSnapshot): Track? {
+      Log.d("Firestore", "DocumentSnapshot is $document")
```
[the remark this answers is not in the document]
KlaKalma: Nope, deleted thanks, was for testing purposes

app/src/main/java/ch/epfl/cs311/wanderwave/model/data/ProfileTrackAssociation.kt
```diff
@@ -9,9 +9,9 @@ package ch.epfl.cs311.wanderwave.model.data
  * @since 2.0
  * @last update 2.0
  */
-data class ProfileTrackAssociation(val profile: Profile, val track: Track) {
+data class ProfileTrackAssociation(val profile: Profile? = null, val track: Track) {
```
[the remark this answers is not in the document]
KlaKalma: We discussed this with @joriba, we can discuss this tomorrow, but for many reasons this is a good idea. Even if this doesn't seem to be the case

app/src/androidTest/java/ch/epfl/cs311/wanderwave/model/ProfileConnectionTest.kt
```diff
@@ -2,30 +2,47 @@ package ch.epfl.cs311.wanderwave.model
...
+    var description: String,
+    var numberOfLikes: Int,
+    var isPublic: Boolean,
+    var profilePictureUri: Uri? = null,
+    var spotifyUid: String,
+    var firebaseUid: String,
+    var topSongs: List<Track> = emptyList(),
+    var chosenSongs: List<Track> = emptyList(),
 )
  */
```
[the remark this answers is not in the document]
KlaKalma: This is the profile data class, just to get an idea of what are the parameters of the profile, it has been here since the begining, and I don't think it's a bad thing. But I can delete it.

app/src/androidTest/java/ch/epfl/cs311/wanderwave/model/ProfileConnectionTest.kt
```diff
@@ -128,4 +148,173 @@ public class ProfileConnectionTest {
...
+    verify { trackConnection.addItemsIfNotExist(profile.chosenSongs) }
+  }
+
+  @Test
+  fun testGetItemCallsOtherGetItem() {
+    val itemId = "testItemId"
+
+    // Mock the ProfileConnection
+    val profileConnection =
+        spyk(ProfileConnection(trackConnection = TrackConnection()), recordPrivateCalls = true)
```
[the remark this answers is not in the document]
KlaKalma: Changed it to a mock.

app/src/main/java/ch/epfl/cs311/wanderwave/model/repository/FirebaseRepository.kt
yzueger: But still, when you need to be able to do something on success, you should just be able to await the result, do whatever the onSuccess would do with the result, and pass it on to another flow if necessary, no?

KlaKalma: No because you don't have all the information in the document snapshot and namely the references to other documents are not stored in the flow of any of the data classes.

yzueger (approved):
All changes addressed, lgtm! Hopefully the coverage is enough :)

PR #257, opened 2024-05-08
Solve the naviagation bug + quality of life update

mbouaissi:
This PR aims to : 

1.  Solve a bug about the edit profile screen, where when the user click on cancel or save, the navigation wasn't working. 
2. Improve the "quality of life". So now, when a user select a picture, he can directly see it, and decide if we want's to save his profile or cancel the modification. (Before the user needed to save his profile and go back to the profile screen so he could see the new image)


mbouaissi:
I haven't really work on the media player, so I'm not an expert about it. Maybe we can create an issue about this problem, and handle it for next sprint

KlaKalma:
> I haven't really work on the media player, so I'm not an expert about it. Maybe we can create an issue about this problem, and handle it for next sprint

Sure

mbouaissi:
Thanks for your review!

PR #259, opened 2024-05-09
End to end test: login and logout

### LINKED ISSUE #170: End-to-end test: Login then log out

yzueger:
Closes #170


app/src/main/java/ch/epfl/cs311/wanderwave/di/AppModule.kt
KlaKalma, line 1: This seems to be refactoring, was this also part of the PR ?

app/src/androidTest/java/ch/epfl/cs311/wanderwave/ui/ProfileTest.kt
```diff
@@ -33,12 +35,15 @@ import org.junit.Test
...
 
 @RunWith(AndroidJUnit4::class)
+@HiltAndroidTest
 class ProfileTest : TestCase(kaspressoBuilder = Kaspresso.Builder.withComposeSupport()) {
 
   @get:Rule val composeTestRule = createAndroidComposeRule<TestActivity>()
 
   @get:Rule val mockkRule = MockKRule(this)
 
+  @get:Rule val hiltRule = HiltAndroidRule(this)
```
KlaKalma, line 45: Why do you need this every time ?

KlaKalma (approved):
Code is nice, but there is so much refactoring and code cleaning that the name of the PR shouldn't be only end to end test. 
Also I thought that we would do the end to end test with the actual spotify api etc. I'm probably wrong.
Thanks for that

app/src/main/java/ch/epfl/cs311/wanderwave/di/AppModule.kt
yzueger: Yes, to be able to do end-to-end tests, we need to mock the modules injected by Hilt. Since we need to create a replacement module for all existing modules we want to change, it's nicer if they are bundled together instead of many classes with a single function. Doesn't change any functionality otherwise, as they are all just installed in the singleton component.

app/src/androidTest/java/ch/epfl/cs311/wanderwave/ui/ProfileTest.kt
yzueger: To be able to use Hilt in the test at all, I had to change the TestRunner to a CustomTestRunner globally. For tests that use a specific activity like `createAndroidComposeRule<TestActivity>()` instead of just `createComposeRule()`, this requires the test to be annotated with @HiltAndroidTest and have the `HiltAndroidRule` as is the case here.
The whole thing is a bit of refactoring, but it all seems to be necessary for end-to-end tests to be possible.

KlaKalma (approved):
Good thanks

PR #278, opened 2024-05-10
Feature/select song screen/main

mbouaissi:
Just merge with main, has already been reviewed before


IB-12 (approved):
already had reviewed this code, still lgtm

PR #281, opened 2024-05-11
Feature/place beacons arround user/main

mbouaissi:
So this is a draft PR. Now we are suppose to see the beacons on the map generated near the user. It's a draft PR because I have some uncertainty and need some advice. Just so we all have the same understanding, this is what i have done: 
1. We look arround the user to see if there is some poi. If it's the case, we place beacons there.
2. If there is not enough poi, we place beacons randomly

 
@aymanbakiri @KlaKalma : 
1. So for adding the beacons on firebase, what I have done in the case of the POI is that i gave as ID the name of the poi. What should i do for the beacons placed randomly, which ID should i pass? I know there is the addItemAndGetId, but it's not implemented in the BeaconConnection, idk if it's normal?
2. Is there a way to maybe re-arrange the database so that beacons are arranged by small region, because i feel like it will be a lot of processing in the future for the algorithm with a lot of beacons

@joriba : 
1. I have encounter a "bug", which was in Map screen, arround line 70. Basically, the getLastKnownLocation was always returning null, and the camera couldn't center on the user, but he was still appearing. Do you have an idea why it does that, and also do you maybe have an alternative way to recover the location of the user?

@IB-12 : 
1. So as you can see, I have modified quite a lot the functions of the placement. I just have a question about the POI functions. I don't really see where you give the argument where you want to recover the beacons of POI from a certain radius. Because I feel that now  the radius is quite small, maybe we could increase it.

@everyone : 
1. I don't know if it's my phone, but now that we add a lot of beacons, the application is quite laggy at the beginning. I don't think I'm doing blocking operation, as I have used _viewModelScope.launch_ . I think the problem lies in the algorithm of placing the beacons randomly, so maybe we can dump it, as I feel like we already have enough beacons with the POI? 

(Also please for now, don't lose your time reviewing the syntax code, I just want advice for the different points mentioned, thx! )


joriba:
@mbouaissi yeah, so about the LastKnownLocation:
This is not the function that is called to update the user's position on the map (this is done by the FastLocationSource that is directly passed to the GoogleMap). getLastKnowLocation is only used to center the camera on the user. The reason I used a separate function for that is that we only want to do that once, when the map is loaded... The user should be able to pan around the map without the camera position constantly changing to his current location. So the function just returns the last known location from the device's cache (I think). 

As for your issue: yeah I noticed that the location returned is not very accurate... but are you sure it's null? At least for me, the location the camera points to seems to change when I move and then open the app again (but it does take some time to register).

mbouaissi:
> @mbouaissi yeah, so about the LastKnownLocation: This is not the function that is called to update the user's position on the map (this is done by the FastLocationSource that is directly passed to the GoogleMap). getLastKnowLocation is only used to center the camera on the user. The reason I used a separate function for that is that we only want to do that once, when the map is loaded... The user should be able to pan around the map without the camera position constantly changing to his current location. So the function just returns the last known location from the device's cache (I think).
> 
> As for your issue: yeah I noticed that the location returned is not very accurate... but are you sure it's null? At least for me, the location the camera points to seems to change when I move and then open the app again (but it does take some time to register).

Thank you for the explanation I will look into it. For the issue, it wasn't always null, it was that sometimes when I launch the application it was null, some other time it was something, I don't know what could cause this situation, maybe I didn't have enough network . But for the location, I think it's quite accurate :), in a sense i have made a tour of EPFL, and the position was good enough imo

joriba:
@mbouaissi so what happens when the location returned is null? Does the camera just not zoom in at all?

mbouaissi:
Yes, that just what happens, my location was shown on the map, but I had to manually zoom on it

IB-12:
@mbouaissi we've already discussed this IRL but if you have any further questions feel free to reach out

app/src/main/java/ch/epfl/cs311/wanderwave/model/utils/BeaconPlacer.kt
```diff
@@ -71,12 +120,18 @@ fun computeDistanceBetweenBeacons(newBeacons: MutableList<Beacon>, beacons: List
...
+    location: Location,
+    newBeacons: MutableList<Beacon>,
+    it: Int,
+    beaconRepository: BeaconRepository
+) {
   val randomLocation = randomLatLongFromPosition(location, BEACON_RADIUS)
   val newBeacon =
       Beacon(
-          id = "beacon${newBeacons.size + it}", // TODO: modify the ID!!!!!!!!
+          id = "", // TODO a modifier
```
IB-12, line 132: Just want to make sure you haven't forgotten this TODO :)

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/MapViewModel.kt
```diff
@@ -29,18 +39,40 @@ constructor(val locationSource: LocationSource, private val beaconRepository: Be
...
+  private fun retrieveBeacons(location: Location1, context: Context) {
+    viewModelScope.launch {
+      // Update _uiState to reflect a loading state
+      _uiState.value = BeaconListUiState(loading = true)
+
+      createNearbyBeacons(
+          location, _beaconList, 10000.0, context, beaconRepository, viewModelScope) {
+            val updatedBeacons =
+                _beaconList
+                    .value // +  placeBeaconsRandomly(_beaconList.value, location, beaconRepository)
```
IB-12, line 69: Is this comment still relevant ? If yes I think it would be good to add that line, else just remove it please.

IB-12 (approved):
Well done, your code is very clear and efficient, I only had some very minor remarks about some comments you left but nothing that prevents it from being merged 👍

app/src/main/java/ch/epfl/cs311/wanderwave/model/utils/BeaconPlacer.kt
mbouaissi: I forgot it thx

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/MapViewModel.kt
mbouaissi: not relevant for now, i will remove it 🫡

app/src/main/java/ch/epfl/cs311/wanderwave/model/utils/BeaconPlacer.kt
```diff
@@ -94,16 +213,35 @@ fun findRandomBeacon(location: Location, newBeacons: MutableList<Beacon>, it: In
...
+ * @last update 2.0
+ */
+fun hasEnoughBeacons(userPosition: Location, beacons: List<Beacon>): Boolean {
+  var nearbyBeacons = mutableListOf<Beacon>()
+  beacons.forEach {
+    if ((userPosition.distanceBetween(it.location)) < BEACON_RADIUS) {
+      nearbyBeacons += it
+    }
+  }
+  return nearbyBeacons.size == BEACON_COUNT
```
IB-12, line 242: shouldn't it be `>=` ?

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/MapViewModel.kt
IB-12: I think you forgor to remove it

IB-12 (changes requested):
Thanks for the changes, there is still a minor issue that I think should be addressed, but otherwise good job 👍

IB-12 (approved):
Thanks for the changes, LGTM

PR #284, opened 2024-05-11
Use dependency injection for the firebase database

KlaKalma:
The goal of this PR is to inject Firebase instead of instantiating it every time like we were doing.
This mainly concerns testing and the different connection files


IB-12 (approved):
Clean code, easy to read, well done

PR #285, opened 2024-05-11
Feature/web api/recover liked songs function

### LINKED ISSUE #267: function to recover the liked songs

mbouaissi:
Now, the user is able to go to his profile, select if he wants to see his top songs or liked songs, and add them to the playlist. This PR also refactor a little bit the code

Closes #267

### DIFF
--- a/app/src/main/java/ch/epfl/cs311/wanderwave/model/spotify/SpotifyController.kt
+++ b/app/src/main/java/ch/epfl/cs311/wanderwave/model/spotify/SpotifyController.kt
@@ -2,6 +2,7 @@ package ch.epfl.cs311.wanderwave.model.spotify
 
 import android.content.Context
 import android.util.Log
+import androidx.lifecycle.viewModelScope
 import ch.epfl.cs311.wanderwave.BuildConfig
 import ch.epfl.cs311.wanderwave.model.auth.AuthenticationController
 import ch.epfl.cs311.wanderwave.model.data.Track
@@ -10,6 +11,7 @@ import com.spotify.android.appremote.api.Connector
 import com.spotify.android.appremote.api.ContentApi
 import com.spotify.android.appremote.api.SpotifyAppRemote
 import com.spotify.android.appremote.api.error.NotLoggedInException
+import com.spotify.protocol.types.ImageUri
 import com.spotify.protocol.types.ListItem
 import com.spotify.protocol.types.PlayerState
 import com.spotify.sdk.android.auth.AuthorizationRequest
@@ -23,9 +25,12 @@ import kotlinx.coroutines.Job
 import kotlinx.coroutines.channels.awaitClose
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.callbackFlow
+import kotlinx.coroutines.flow.firstOrNull
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
+import org.json.JSONObject
 
 class SpotifyController
 @Inject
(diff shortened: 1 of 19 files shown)


app/src/main/java/ch/epfl/cs311/wanderwave/model/spotify/SpotifyController.kt
```diff
@@ -315,3 +319,100 @@ constructor(
...
+}
+
+/**
+ * Get all the element of the main screen and add them to the top list
+ *
+ * @author Menzo Bouaissi
+ * @since 2.0
+ * @last update 3.0
+ */
+fun retrieveChildFromSpotify(
```
IB-12, line 351: This function and the one above look extremely similar, do you think you could find a way to "merge" them to avoid duplication ? It's really not a big deal but it'd be nice to have

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/tracklist/TrackList.kt
```diff
@@ -46,16 +51,19 @@ fun TrackList(
         horizontalArrangement = Arrangement.SpaceBetween) {
           if (title != null) {
             Text(
-                text = title,
+                text = title.toLowerCase().replace("_", " "),
```
IB-12, line 54: Do we really want to have all songs titles in lower case ? I think we should try to stick as close as possible to the way Spotify displays it

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/SelectSongScreen.kt
```diff
@@ -56,26 +68,24 @@ fun SelectSongScreen(navActions: NavigationActions, viewModel: ProfileViewModel)
...
+          items(displayedList, key = { it.id }) { listItem ->
+            TrackItem(
+                listItem,
+                onClick = {
+                  if (listItem.hasChildren) {
+                    viewModel.retrieveChild(listItem)
+                  } else {
+                    viewModel.addTrackToList(
+                        isTopSongsListVisible?.let {
+                          if (it) ListType.TOP_SONGS else ListType.LIKED_SONGS
```
IB-12, line 82: Careful with the nested ifs here, they will negatively impact our complexity

IB-12 (changes requested):
Very good code as usual, the only issue I saw was a nested if which I think we should really avoid given M3's requirements. Other than that I had a couple of minor suggestions that you can feel free to disregard if you disagree with them.

app/src/main/java/ch/epfl/cs311/wanderwave/model/spotify/SpotifyController.kt
mbouaissi: Actually, they are not the same function. As you can see, one is calling  getAllElementFromSpotify and the other one getAllChildren(item). I can try to modularize the fact of adding element into the list. If you see a better way of merge it you can show me

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/tracklist/TrackList.kt
mbouaissi: Actually, if you look at how the component tracklist is called, we do not modify the title of the song here, but the title of the track list (TOP_SONGS => top songs, LIKED_SONGS=> liked songs). I can maybe change that, but I think that it will be change to a way better maneer when we are going to look at the UI again, don't worry about it 😉

IB-12: You're right, simple misunderstanding on my part. Thanks for clarifying !

aymanbakiri (changes requested):
Your code is good ,very clear and well commented. Also you have done fairly exhaustive tests.
Now the user can go to his profile in order to decide if he wants to see his top songs or his liked songs. The user also have the possibility to add these songs to the playlist.

I think that small change that you need to do is to remove the nested if in selectSongScreen in order to not increase complexity
Good job , thanks

aymanbakiri (approved):
Thanks for your work ,I think it is ready to merge

IB-12 (approved):

PR #286, opened 2024-05-11
add item and get id now add the tracks in beacon connection

KlaKalma:
@mbouaissi 
This PR add the fact that when retrieving the ID, the tracks that are in the beacon are also added to firebase.


mbouaissi:
LGTM, thanks for doing it :+1:

mbouaissi (approved):
lgtm

PR #287, opened 2024-05-12
improved UI test + some aesthetics

### LINKED ISSUE #165: Add missing UI tests

tsogtb:
closes #165 
In this PR, 

- player screens (mini, exclusive) were properly tested
- loading icon is now aligned with the login icon and now uses different colors for better aesthetics
- exclusive player drag handle switch button is now properly placed and the color is more vibrant


app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/player/ExclusivePlayer.kt
```diff
@@ -72,10 +79,10 @@ fun SwitchComponent(checked: MutableState<Boolean>) {
       onCheckedChange = { checked.value = it },
       colors =
           SwitchColors(
-              checkedThumbColor = Color.White,
+              checkedThumbColor = pink,
```
KlaKalma, line 82: Those colors maybe should be in the ressource folder, so we can decided the theme like that

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/login/WelcomeTitle.kt
```diff
@@ -8,27 +8,32 @@ import androidx.compose.material3.Text
...
-import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.style.TextAlign
-import ch.epfl.cs311.wanderwave.R
 import ch.epfl.cs311.wanderwave.ui.components.animated.AnimatedAppIcon
 
 @Composable
-fun WelcomeTitle(modifier: Modifier) {
+fun AppLogoWithTitle(
+    modifier: Modifier,
+    initialColor: Color,
```
KlaKalma, line 19: Same here, the colors should be in the ressource folder (in the theme), you can very easily add material colors. Especially as we need to have dark and light theme at the end I think. It will be much easier like that.

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/utils/LoadingScreen.kt
```diff
@@ -1,17 +1,32 @@
...
       modifier = modifier.fillMaxSize().testTag("loadingScreen"),
-      contentAlignment = androidx.compose.ui.Alignment.Center) {
-        AnimatedAppIcon()
+      contentAlignment = Alignment.Center) {
+        Column {
+          Box(modifier = Modifier.weight(1f))
+          AppLogoWithTitle(
+              modifier = Modifier.weight(4f),
+              initialColor = coldStart,
+              finalColor = coldEnd,
```
KlaKalma, line 23: Same here

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/utils/LoadingScreen.kt
```diff
@@ -1,17 +1,32 @@
...
+              title = "",
+              subTitle = "")
+          Box(modifier = modifier.weight(1f))
+        }
       }
 }
+
+val coldStart = Color(0xFF03045E)
+val coldEnd = Color(0xFFCAF0F8)
```
KlaKalma, line 32: This as well

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/LoginScreen.kt
```diff
@@ -54,7 +57,12 @@ fun LoginScreen(
 
   Column(modifier = Modifier.testTag("loginScreen")) {
     LoginAppLogo(modifier = Modifier.weight(1f))
-    WelcomeTitle(modifier = Modifier.weight(4f))
+    AppLogoWithTitle(
+        modifier = Modifier.weight(4f),
+        initialColor = MaterialTheme.colorScheme.primary,
+        finalColor = Color.Red,
```
KlaKalma, line 63: The red should be in the theme, because we should keep the app color scheme consistent.

KlaKalma (approved):
Thanks for that, there are still some little comments I made about the differents things that may be improved. Mainly putting everything in the theme instead of hardcoding the different colors.

KlaKalma (changes requested):
Thanks for that, there are still some little comments I made about the differents things that may be improved. Mainly putting everything in the theme instead of hardcoding the different colors.

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/LoginScreen.kt
tsogtb: I think its not really part of the theme but rather a fixed color in this case

mbouaissi (approved):
lgtm

PR #289, opened 2024-05-12
delete add profiles if not exist

KlaKalma:
This PR delete the function we created for testing for the beacon connection that add a profile if it doesn't exist, now as the profile in the track profile association can be null, we don't have any problems any more. And as yoric is going to connect the profiles to the function, it should work flowlessly.


IB-12 (approved):
As usual your code is clean and easy to understand, well done.

PR #291, opened 2024-05-13
Get CI log as artifact

KlaKalma:
The goal of this PR is to get the logs for the CIs that didn't work, the logs will only be for the instrumentation test.


mbouaissi (approved):
lgtm

PR #292, opened 2024-05-13
Feature/link profile to auth/persist profile

yzueger:
Creates the user profile on Firestore if it doesn't exist yet, and loads it otherwise

### DIFF
--- a/app/src/main/java/ch/epfl/cs311/wanderwave/model/remote/ProfileConnection.kt
+++ b/app/src/main/java/ch/epfl/cs311/wanderwave/model/remote/ProfileConnection.kt
@@ -26,14 +26,11 @@ class ProfileConnection(
   override val db = database
   private val coroutineScope = CoroutineScope(Dispatchers.IO + SupervisorJob())
 
-  override fun isUidExisting(spotifyUid: String, callback: (Boolean, Profile?) -> Unit) {
+  override fun isUidExisting(firebaseUid: String, callback: (Boolean, Profile?) -> Unit) {
     db.collection("users")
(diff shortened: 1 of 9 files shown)


app/src/main/java/ch/epfl/cs311/wanderwave/model/remote/ProfileConnection.kt
```diff
@@ -26,14 +26,11 @@ class ProfileConnection(
...
+  override fun isUidExisting(firebaseUid: String, callback: (Boolean, Profile?) -> Unit) {
     db.collection("users")
-        .whereEqualTo("spotifyUid", spotifyUid)
+        .document(firebaseUid)
         .get()
-        .addOnSuccessListener { documents ->
-          val isExisting = documents.size() > 0
-          callback(isExisting, if (isExisting) documentToItem(documents.documents[0]) else null)
-        }
+        .addOnSuccessListener { document -> callback(true, documentToItem(document)) }
```
KlaKalma, line 33: Like we sais we should change this in the future, I don't know if you want to do it now, in your branch or for a future issue.

KlaKalma (approved):
Good, I just left a comment about the isUIDExisting function. If you want to do it later just merge.

app/src/main/java/ch/epfl/cs311/wanderwave/model/remote/ProfileConnection.kt
yzueger: I'll wait for the error handling in the FirebaseConnection, that way we don't need this function in the first place.

PR #294, opened 2024-05-14
 save the profile in firebase

aymanbakiri:
I have edited the `EditProfileScreen` so that when a profile is created, it is saved to Firebase. This ensures that the next time you connect to the app, you can load the saved profile.


IB-12 (approved):
Code is clean and well commented, LGTM 👍

IB-12 (approved):
Still LGTM

PR #296, opened 2024-05-14
Updated the test

IB-12:
I've adapted the test to the new flow and fixed some minor issues


mbouaissi (dismissed):
Thx for the change. I just have to remarks:
1. pls put your file in the same directory as the other end to end tests
2. Regarding your problem with the CI, maybe it's on line 80, try this instead :   val dispatcher = UnconfinedTestDispatcher()

IB-12:
> Thx for the change. I just have to remarks:
> 
> 1. pls put your file in the same directory as the other end to end tests
> 2. Regarding your problem with the CI, maybe it's on line 80, try this instead :   val dispatcher = UnconfinedTestDispatcher()

Thank you for your review. The issue was that I forgot to release the `Intent` in my end-to-end test, but it's solved now. I've also put the test in the correct directory.

Issue #262, opened 2024-05-10
Connect the auth to firebase, and retrieve the profile


yzueger:
We need to have a way to have my profile and modify it

PR #298, opened 2024-05-15
Revert " save the profile in firebase"

yzueger:
Reverts G12-Wanderwave/Wanderwave#294

This conflicts with #292 which was already merged into the feature branch `feature/link-profile-to-auth` before, and implements just a subset of it.

### DIFF
--- a/app/src/androidTest/java/ch/epfl/cs311/wanderwave/viewmodel/ProfileViewModelTest.kt
+++ b/app/src/androidTest/java/ch/epfl/cs311/wanderwave/viewmodel/ProfileViewModelTest.kt
@@ -1,9 +1,7 @@
 package ch.epfl.cs311.wanderwave.viewmodel
 
-import androidx.compose.material3.SnackbarHostState
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import ch.epfl.cs311.wanderwave.model.data.ListType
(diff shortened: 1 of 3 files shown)


aymanbakiri (changes requested):
I see that there was a misunderstanding when we split the tasks. You are doing the following task #262: Connect the auth to Firebase and retrieve the profile, and I am doing the following task #263: Create a profile screen.

Personally, what I understand is to change the edit profile screen so that the user can create a profile and save it into Firebase. This way, when the user connects again, they can load their profile after logging into the app. But this is what you were also asked to do.

So, do you think that I can create a screen that allows a user to create their profile and save it into Firebase using what you have already implemented?

yzueger:
As per Discord:

For your PR I guess it would make sense to have an EditProfile screen where you can't cancel it but have to enter the details. Once Clarence is done updating the Firebase Connections, it should also be easy to detect if it is the first time and automatically redirect the user to that screen in that case. I'm also waiting for these changes at the moment.

And please do PRs to merge into the `feature/link-profile-to-auth/main` branch for this, to avoid merge conflicts. There, the saving functionality is already implemented

yzueger:
Is there any reason not to revert the PR, or what changes are you requesting?

aymanbakiri (approved):
Ok

PR #299, opened 2024-05-15
Fix the progress bar UI

IB-12:
The progress bars in both media players are now in sync with the actual song


aymanbakiri (approved):
Your code is good ,very clear and well commented even if it is small. 
Thanks for synchronizing the progress bars in both media players with the actual song
UI looks good now.
Good job , thanks

PR #302, opened 2024-05-15
show recently played tracks on the tracklistScreen when toggled

joriba:
When listening to a song, the spotifyController now remembers it and adds the song to a list.
On the track list screen, one can toggle the switch to show the songs recently listened to. This list is also playable, so we can listen to recent tracks from there.


aymanbakiri (approved):
Your code is good ,very clear and well commented even if it is small.
Now, the spotifyController is able to remember a song and add it to the list
Also, in the track list screen we can use the switch button to show the songs recently listened .
Good job , thanks

PR #318, opened 2024-05-16
Feature/web api/main

IB-12:
This will merge all the various features related to the web API into main

- #265 
- #266 
- #267 
- #268

PR #334, opened 2024-05-16
Adapt the add track button in the beaconscreen

IB-12:
This allows users to manually add a track to a beacon.
Will be fully functional once #318 is merged


mbouaissi (approved):
lgtm, just a question, with your current implementation, do you think it's going to be simple to drop a random songs from the user list instead of the first song? Otherwise thx for the nice code

IB-12:
> lgtm, just a question, with your current implementation, do you think it's going to be simple to drop a random songs from the user list instead of the first song? Otherwise thx for the nice code

You mean when passively dropping song ? Yeah I don't think it'll be too much work to change that, Thank you for your review

PR #337, opened 2024-05-16
Fixed the issue in BeaconScreen

IB-12:
This allows users to add songs to a beacon from the BeaconScreen


joriba (approved):
Okay, I mean implementing the interface where half of the functions are not needed is kind of ugly... but I guess I'll greenlight it since we need the fix quickly.
Maybe if there's a way to do it a bit cleaner, we can do that for the next milestone?

mbouaissi (approved):
LGTM, thanks for making possible to drop songs to beacon

Issue #198, opened 2024-04-25
As a user, I want to be presented with a high quality, aesthetically pleasing UI/UX and a Responsive Design so that I can use the app on multiple devices and enjoy the experience.


KlaKalma:
* [x] #201 
* [x] #206 
* [x] #207

PR #339, opened 2024-05-17
Launcher Icon + AppBottomBar update + (Edit)ProfileScreen UI + TrackList UI fix

tsogtb:
revisits #198 
In this PR,

- App launcher Icon was set.
- Launcher Background is Black.
- (Edit) Profile Screen Buttons are a bit more rectangular and different color (EditProfileScreen now fills max height)
- AppBottomBar container is Black, Map Icon is now Beacon Icon, and other 2 destinations have small label
- On TrackList Screen, the last Item of the LazyList was hard to select because of the miniPlayer and this was padded.
- Made some room for song picture
- In Profile Screen there was no wrapper around the column and row so they were overlapping hence failing the test on smaller screen in the CI. It got fixed.


app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/AppBottomBar.kt
```diff
@@ -45,34 +55,49 @@ fun AppBottomBar(navActions: NavigationActions) {
...
+                  }
+            }
 
-    Spacer(Modifier.weight(1f)) // Spacer for centering the middle icon
+        IconButton(
+            onClick = { navActions.navigateToTopLevel(TOP_LEVEL_DESTINATIONS[1].route) },
+            modifier = Modifier.weight(1f).testTag("bottomAppBarButton" + Route.MAP.routeString)) {
+              Image(
+                  modifier = Modifier.weight(4f).padding(top = 5.dp, bottom = 5.dp),
+                  painter = painterResource(id = R.drawable.beaconlogo),
```
joriba, line 82: I like the beacon logo here instead of the default map marker. Is there maybe a way to make it clearer that this is the map screen? Maybe a map icon with the beacon logo on it?

joriba (changes requested):
Had some comments about the UI and code.

app/src/main/ic_launcher-playstore.png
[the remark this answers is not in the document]
mbouaissi: Me too

mbouaissi (changes requested):
Thanks for the code, i just have a few questions:
1. Why doing the "UI" part know, because the "backend" part is not done yet, and it will probably change, so we will have to do the UI again later.
2. Why is the loading screen at the beginning with a new icon, isn't it better to keep the same color scheme for all the logo
3. Is it possible to make the media player disappear when no songs is playing
4. Dont forget to link to which issue is your pr related

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/ProfileScreen.kt
```diff
@@ -88,7 +99,16 @@ fun ProfileScreen(navActions: NavigationActions, viewModel: ProfileViewModel) {
...
+            modifier = Modifier.testTag("toggleSongList"),
+            colors =
+                ButtonColors(
+                    contentColor = Color.Black,
+                    containerColor =
+                        if (isTopSongsListVisible) MaterialTheme.colorScheme.primary
+                        else spotify_green,
+                    disabledContentColor = Color.Gray,
+                    disabledContainerColor = Color.Black),
+            shape = RoundedCornerShape(size = 10.dp)) {
```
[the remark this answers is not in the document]
tsogtb: I agree but but it is somehow unclear how we can define one since I didn't wanna touch what was already there and as @mbouaissi mentioned the UI will come later when the Backend is done. So I am guessing this can be tolerated for now?

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/AppBottomBar.kt
tsogtb: I think it is somehow easily understood once the user sees beacons on the map.

tsogtb: But I will try to come up with something that doesn't look too crowded

tsogtb:
> Thanks for the code, i just have a few questions:
> 
>     1. Why doing the "UI" part now, because the "backend" part is not done yet, and it will probably change, so we will have to do the UI again later.
> 
>     2. Why is the loading screen at the beginning with a new icon, isn't it better to keep the same color scheme for all the logo
> 
>     3. Is it possible to make the media player disappear when no songs is playing
> 
>     4. Dont forget to link to which issue is your pr related

1. If you were referring to the Profile screen there isn't much change otherwise can you clarify which UI part you are referring to.
2. Agreed changed it with the primary color old Icon.
3. I can do that with another PR.
4. Thanks for the tip, I linked them to #198  as these don't really add any functionality. Correct me if there could be something else.

And thanks for the comment

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/ProfileScreen.kt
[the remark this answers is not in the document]
joriba: Yeah I guess it's fine

app/src/main/java/ch/epfl/cs311/wanderwave/ui/components/AppBottomBar.kt
joriba: thanks

tsogtb: image
turns out the layout doesn't allow me to have vertical space and overlapping makes it look bad. I propose this version lmk what you think

app/src/main/res/drawable/logout_icon.xml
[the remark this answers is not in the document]
tsogtb: I can't seem to find one in the Icons.Default

joriba (approved):
Lgtm

PR #340, opened 2024-05-17
Comment the beacon logo

mbouaissi:
This PR just comment the part where we set up the logo of the beacons, so that the application is more fluid


IB-12 (approved):
LGTM

PR #346, opened 2024-05-19
Feature/link profile to auth/main

### LINKED ISSUE #262: Connect the auth to firebase, and retrieve the profile

yzueger:
Closes #262 
Feature branch merge


app/src/main/java/ch/epfl/cs311/wanderwave/model/remote/ProfileConnection.kt
```diff
@@ -26,14 +26,11 @@ class ProfileConnection(
   override val db = database
   private val coroutineScope = CoroutineScope(Dispatchers.IO + SupervisorJob())
 
-  override fun isUidExisting(spotifyUid: String, callback: (Boolean, Profile?) -> Unit) {
+  override fun isUidExisting(firebaseUid: String, callback: (Boolean, Profile?) -> Unit) {
```
KlaKalma, line 29: As we said that both are the same we should probably do a refactoring of it soon where we have only one uid no ? And delete the spotify UID. Of course not for this PR

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/ProfileViewModel.kt
```diff
@@ -99,13 +101,26 @@ constructor(
     _isInPublicMode.value = !_isInPublicMode.value
   }
 
-  fun getProfileByID(id: String) {
-    viewModelScope.launch {
-      repository.getItem(id).collect { fetchedProfile ->
+  suspend fun getProfileByID(id: String, create: Boolean) {
+    repository.isUidExisting(id) { exists, fetchedProfile ->
```
KlaKalma, line 105: You probably should use getItem instead of isUidExisting

KlaKalma (commented):
Like I said can you change addTrackToBeacon in the beacon connection file so that the actual profile is used ?

KlaKalma (approved):
LGTM

PR #348, opened 2024-05-19
Fix a bug

mbouaissi:
Fix a bug that was fix before but lost during a merge

### DIFF
--- a/app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/SelectSongScreen.kt
+++ b/app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/SelectSongScreen.kt
@@ -220,6 +220,12 @@ fun handleItemClick(
     viewModel.retrieveChild(listItem)
     return
   }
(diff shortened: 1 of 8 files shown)


yzueger (approved):
lgtm

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/SelectSongScreen.kt
```diff
@@ -220,6 +220,12 @@ fun handleItemClick(
     viewModel.retrieveChild(listItem)
     return
   }
+  if (listItem.id.contains("spotify:") ||
+      listItem.id.isBlank()) { // TODO: create an issue for handling this
```
IB-12, line 224: Make sure to create this issue if you haven't already done so

mbouaissi: yes chef

PR #349, opened 2024-05-19
Feature/link profile to auth/set user profile in track association

yzueger:
Some cleaning up with the new Firebase functionality, as well as correctly setting the users profile when adding a track to a beacon


KlaKalma (approved):
Lgtm nice

PR #350, opened 2024-05-19
 handling first time a user create profile

aymanbakiri:
I added a `canCancel` option to the `EditProfileScreen` to make cancellation impossible if f it’s the user's first time connecting. In `SpotifyConnectScreen`, I added logic to check if it’s the user's first time connecting by attempting to fetch their profile from the repository. If the profile fetch fails with the error "Document does not exist," the user is redirected to the `EditProfileScreen` instead of the usual flow.
#closes 263


yzueger (changes requested):
I tested it and it does not seem to be working at all. I don't think you can use the ProfileViewModel for this, as that is for a different screen. You should be able to check if the profile exists directly using the ProfileRepository, and the users id you can get from the AuthenticationController using `.getUserData().id`

PR #353, opened 2024-05-19
Add banned to the profile and likes to the beacon

### LINKED ISSUE #310: Have a like counter on firebase for tracks in beacons

### LINKED ISSUE #312: The profile contains a list of banned songs (local and firebase)

KlaKalma:
In this PR, I add the possibility to have a profile with liked and banned songs as well as a like counter on the beacon.

As discussed earlier, I don't store the profiles in the beacon likes as we thought that it would be better to have the view model verify that the user didn't already like a song.

Closes #312 
Closes #310


mbouaissi (commented):
thanks for this first draft, the code lgtm. I just have a question regarding the likes. Why have you put the like in the beacon instead of the profileTrackAssociation class? Because for me when a user likes a song, it would make more sense to show wich tuple (profile, track) was liked, rather than showing which beacon was liked. Moreover, I think it will be easier in the future when we will linked your work with the profile counter

mbouaissi:
Also, shouldn't you merge to feature/place-beacons/main instead of main directly?

KlaKalma:
> thanks for this first draft, the code lgtm. I just have a question regarding the likes. Why have you put the like in the beacon instead of the profileTrackAssociation class? Because for me when a user likes a song, it would make more sense to show wich tuple (profile, track) was liked, rather than showing which beacon was liked. Moreover, I think it will be easier in the future when we will linked your work with the profile counter

You're totally right, thanks for pointing that out, I just was mistaken, I think we can still leave the counter on the beacon, I'll just add it on the tracks.

KlaKalma:
> Also, shouldn't you merge to feature/place-beacons/main instead of main directly?

The branch is 400 commits behind main, it was for a previous feature, I think we can just directly merge into main.

KlaKalma:
@yzueger I modified a bit the profile track association, tell me what you think.

aymanbakiri (dismissed):
Clarence ,another time you have done my task which was store the liked songs on the profile #308. Please remove this part from your code.
Thanks

app/src/androidTest/java/ch/epfl/cs311/wanderwave/model/DataClassesTest.kt
```diff
@@ -168,7 +169,9 @@ class DataClassesTest {
             "isPublic" to true,
             "profilePictureUri" to "https://example.com/image.jpg",
             "chosenSongs" to listOf<DocumentReference>(),
-            "topSongs" to listOf<DocumentReference>())
+            "topSongs" to listOf<DocumentReference>(),
+            "bannedSongs" to listOf<DocumentReference>(),
+            "likedSongs" to listOf<DocumentReference>())
```
yzueger, line 174: This still references the likedSongs, which was removed. Locally this test fails, I'm not sure how it passed in the CI.

app/src/androidTest/java/ch/epfl/cs311/wanderwave/model/ProfileConnectionTest.kt
```diff
@@ -112,7 +109,8 @@ public class ProfileConnectionTest {
             "profilePictureUri" to "https://example.com/image.jpg",
             "topSongs" to listOf<DocumentReference>(),
             "chosenSongs" to listOf<DocumentReference>(),
-        )
+            "bannedSongs" to listOf<DocumentReference>(),
+            "likedSongs" to listOf<DocumentReference>())
```
yzueger, line 113: Also references likedSongs and fails test locally

app/src/main/java/ch/epfl/cs311/wanderwave/model/remote/BeaconConnection.kt
```diff
@@ -77,53 +77,49 @@ class BeaconConnection(
...
+                  trySend(Result.success(updatedBeacon))
+                }
+                result.onFailure { exception ->
+                  trySend(Result.success(beacon))
+                  Log.e("Firestore", "Error getting profile and track: ", exception)
                 }
               }
-            } else {
-              trySend(Result.success(beacon))
-              Log.e("Firestore", "Tracks are not in the correct format ")
```
yzueger, line 122: Looks like a lot of formatting stuff here, just to check is this purely formatting or did I miss an actual change past the `coroutineScope.launch`?

app/src/main/java/ch/epfl/cs311/wanderwave/model/remote/BeaconConnection.kt
```diff
@@ -144,29 +140,16 @@ class BeaconConnection(
...
-            "tracks" to
-                beacon.profileAndTrack.map { profileAndTrack ->
-                  hashMapOf(
-                      "creator" to
-                          db.collection("users")
-                              .document(profileAndTrack.profile?.firebaseUid ?: ""),
-                      "track" to db.collection("tracks").document(profileAndTrack.track.id))
-                })
-    return beaconMap
-  }
```
yzueger, line 161: Since the toMap is firebase specific, I would still prefer these methods in the firebase part and not the dataclasses, but probably for another PR.

app/src/main/java/ch/epfl/cs311/wanderwave/model/remote/BeaconConnection.kt
```diff
@@ -175,42 +158,29 @@ class BeaconConnection(
...
-                })
+                associations
+                    .map { it.toMap(db) }
+                    .toMutableList()
+                    .apply {
+                      val trackRef =
+                          db.collection(trackConnection.collectionName).document(track.id)
+                      add(
+                          hashMapOf(
+                              "profile" to profileRef, "track" to trackRef, "numberOfLikes" to 0))
```
yzueger, line 179: Looks great like this, much better than before

app/src/main/java/ch/epfl/cs311/wanderwave/model/remote/FirebaseConnection.kt
```diff
@@ -95,6 +95,7 @@ abstract class FirebaseConnection<T, U>(open val db: FirebaseFirestore) {
               documentToItem(document)?.let {
                 // The document transform function is used when references are inside and need to be
                 // fetched
+                Log.d("Firestore", "DocumentSnapshot data inside getItem: ${document.data}")
```
yzueger, line 98: I would probably not leave that in for every single getItem call.

app/src/main/java/ch/epfl/cs311/wanderwave/model/remote/ProfileConnection.kt
```diff
@@ -59,80 +59,77 @@ class ProfileConnection(
...
+        // map to a list of track
+        ?.fold(flowOf(Result.success(listOf<Track>()))) { acc, track ->
+          acc.combine(track) { accTracks, track -> accTracks.map { tracks -> tracks + track } }
+        } ?: flowOf(Result.failure(Exception("Could not retrieve topSongs")))
+    // reduce the lists of flows to a
+    // single flows that contains the
+    // list of tracks
+  }
+
+  fun isValidObject(obj: Any?): Boolean {
```
yzueger, line 132: This function name is not very intuitive, it seems to check that something is a list, but mentions object in the name. Renaming or some documenting comments could help.
Also, maybe just returning the correctly typed value or `null` if not might be nice.

app/src/main/java/ch/epfl/cs311/wanderwave/model/remote/ProfileConnection.kt
```diff
@@ -59,80 +59,77 @@ class ProfileConnection(
...
+            updatedProfile
+                .map { Result.success(it) }
+                .collect { result ->
+                  result.onSuccess { profile -> trySend(Result.success(profile)) }
+                }
           }
+          // } else {
+          //   Log.e("ProfileConnection", "Tracks have wrong firebase format")
+          //   trySend(Result.success(profile))
+          // }
```
yzueger, line 109: Commented code

yzueger (changes requested):
2 tests don't pass locally, but somehow seem to have passed on CI. Otherwise I just have some nitpicks.
The Firebase stuff starts looking quite nice with this.

app/src/main/java/ch/epfl/cs311/wanderwave/model/remote/BeaconConnection.kt
KlaKalma: There is no actual change it's just that the document cannot be null at this state so I just put a !! instead of this ?.let.

app/src/main/java/ch/epfl/cs311/wanderwave/model/remote/BeaconConnection.kt
KlaKalma: We would need another file for that I think, or maybe put it in some kind of subclass of the firebase connection

app/src/main/java/ch/epfl/cs311/wanderwave/model/remote/ProfileConnection.kt
KlaKalma: Very nice idea !

app/src/androidTest/java/ch/epfl/cs311/wanderwave/model/ProfileConnectionTest.kt
KlaKalma: It passes for me, I don't know why, either way I deleted

yzueger (approved):
lgtm now

PR #354, opened 2024-05-20
Improve beacon icon perf

IB-12:
This PR restores the custom beacon markers we had before which should now have a significantly reduced impact on performance.

Edit: The icon is now generated each time we load the map instead of once for each beacon. Ideally we would generate it only once but I'm still working on that.


mbouaissi:
Thanks for doing this. For not generating the image every can't you do something i did with the based profile picture? Like you save it locally and just reference it. (idk if its possible with the bitmap, i have used a imageurl)

IB-12:
Thanks for the input, unfortunately due to the way the Google Maps API works your trick doesn't work. I have however found a way to generate the icon only once on app launch. The performance is still not as good as the original markers but at least it's usable now.

mbouaissi (approved):
lgtm, I just have an idea for improving the perf. When you look at the image(beaconlogo.png) the size is quite big for a marker(84kB for 1500x1500 pixel) maybe by trying to work on this, and reducing the source image it can changes something. If not it's ready to merge for me, nice code :+1:

KlaKalma:
> lgtm, I just have an idea for improving the perf. When you look at the image(beaconlogo.png) the size is quite big for a marker(84kB for 1500x1500 pixel) maybe by trying to work on this, and reducing the source image it can changes something. If not it's ready to merge for me, nice code 👍

I think this is the main thing, because we are resizing it very small at the end but not changing its resolution it makes things laggy, if you want I can generate a lower res version of the picture.

IB-12:
I don't think it will help with performance because by the time it's displayed on the map, the resolution is already 100x100 due to the resize. But I agree it's an unnecessary operation, even if only performed once at launch.

PR #359, opened 2024-05-21
Automatically retrieve drop songs/main

### LINKED ISSUE #358: retrieve songs by walking near a beacon

mbouaissi:
Now, when the user open the application and switch to another application (or shut down the app), it still runs +  when a user pass near a beacon, we retrieve a random song inside it. This PR doesn't add the song to the tracklist, nor add song from the user playlist to the beacon. 

Closes #358


app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/MapScreen.kt
```diff
@@ -69,40 +78,75 @@ fun MapScreen(navigationActions: NavigationActions, viewModel: MapViewModel = hi
...
+        if (context != null) {
+          viewModel.loadBeacons(context, locationState.value!!)
+        }
+        val tempBeacon = findClosestBeacon(locationState.value!!, viewModel.beaconList.value)
+        if (tempBeacon != null) {
+          viewModel.getRandomSong(tempBeacon.id)
+          Log.d(
+              "MapScreen",
+              "retrieved song: ${song.value}") // TODO: the song is retrieved here if exist
+          // TODO: implement here the fact to drop a song to the nearest beacon
```
IB-12, line 123: Might be worth already creating an issue for this in order not to forget it

IB-12 (changes requested):
Your code is very good, however when I try to launch the app it crashes with the following error:
```
FATAL EXCEPTION: main
Process: ch.epfl.cs311.wanderwave, PID: 15878

java.lang.RuntimeException: Unable to create service ch.epfl.cs311.wanderwave.model.utils.LocationUpdatesService: java.lang.IllegalArgumentException: ch.epfl.cs311.wanderwave: Targeting S+ (version 31 and above) requires that one of FLAG_IMMUTABLE or FLAG_MUTABLE be specified when creating a PendingIntent.

Strongly consider using FLAG_IMMUTABLE, only use FLAG_MUTABLE if some functionality depends on the PendingIntent being mutable, e.g. if it needs to be used with inline replies or bubbles.```

Which is why I can't let you merge this yet unfortunately :)

IB-12:
Looks like I just needed to run a gradle clean, so don't waste your time looking for this non-existent bug

IB-12 (dismissed):
LGTM

IB-12 (approved):
Well done for fixing that, looked quite tricky

PR #361, opened 2024-05-22
Feature/like songs/like locally

### LINKED ISSUE #309: Have a like counter on the track profile association

IB-12:
This allows users to like (and unlike) songs in beacons, which adds them to a dedicated list and increases the song's OP's like counter.

Closes #309

### DIFF
--- a/app/src/main/java/ch/epfl/cs311/wanderwave/model/data/ProfileTrackAssociation.kt
+++ b/app/src/main/java/ch/epfl/cs311/wanderwave/model/data/ProfileTrackAssociation.kt
@@ -21,10 +21,12 @@ data class ProfileTrackAssociation(
   }
 
   fun likeTrack(profile: Profile): ProfileTrackAssociation {
+    this.profile?.numberOfLikes = this.profile?.numberOfLikes!! + 1
     return ProfileTrackAssociation(profile, track, likers + profile, likes + 1)
   }
 
   fun unlikeTrack(profile: Profile): ProfileTrackAssociation {
+    this.profile?.numberOfLikes = this.profile?.numberOfLikes!! - 1
     return ProfileTrackAssociation(profile, track, likers - profile, likes - 1)
   }
 }
--- a/app/src/main/java/ch/epfl/cs311/wanderwave/model/data/ProfileTrackAssociation.kt
+++ b/app/src/main/java/ch/epfl/cs311/wanderwave/model/data/ProfileTrackAssociation.kt
@@ -25,18 +25,18 @@ data class ProfileTrackAssociation(
   }
 
   fun likeTrack(profile: Profile): ProfileTrackAssociation {
-    if (likersId.contains(profile.firebaseUid)) return this
-    if (this.profile != null) {
+    if (this.profile != null && !likersId.contains(profile.firebaseUid)) {
       this.profile.numberOfLikes += 1
+      return ProfileTrackAssociation(profile, track, likersId + profile.firebaseUid, likes + 1)
     }
-    return ProfileTrackAssociation(profile, track, likersId + profile.firebaseUid, likes + 1)
+    return this
   }
 
   fun unlikeTrack(profile: Profile): ProfileTrackAssociation {
-    if (!likersId.contains(profile.firebaseUid)) return this
-    if (this.profile != null) {
+    if (this.profile != null && likersId.contains(profile.firebaseUid)) {
       this.profile.numberOfLikes -= 1
+      return ProfileTrackAssociation(profile, track, likersId - profile.firebaseUid, likes - 1)
     }
-    return ProfileTrackAssociation(profile, track, likersId - profile.firebaseUid, likes - 1)
+    return this
   }
 }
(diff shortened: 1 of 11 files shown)


app/src/main/java/ch/epfl/cs311/wanderwave/model/data/ProfileTrackAssociation.kt
```diff
@@ -9,9 +9,24 @@ package ch.epfl.cs311.wanderwave.model.data
...
+    val likers: List<Profile> = emptyList(),
+    val likes: Int = 0
+) {
 
   fun toMap(): Map<String, Any?> {
     return hashMapOf("profile" to profile?.toMap(), "track" to track.toMap())
   }
+
+  fun likeTrack(profile: Profile): ProfileTrackAssociation {
+    this.profile?.numberOfLikes = this.profile?.numberOfLikes!! + 1
```
KlaKalma, line 24: This seems like a bad null assert, Can you add a test before that test if the profile is not null, 
something like 
`this.profile?.let{...}
?: // case if it's null`

KlaKalma: And maybe add a safety check that the profile is not already in the likers.

app/src/main/java/ch/epfl/cs311/wanderwave/model/data/ProfileTrackAssociation.kt
```diff
@@ -9,9 +9,24 @@ package ch.epfl.cs311.wanderwave.model.data
...
+    return ProfileTrackAssociation(profile, track, likers + profile, likes + 1)
+  }
+
+  fun unlikeTrack(profile: Profile): ProfileTrackAssociation {
+    this.profile?.numberOfLikes = this.profile?.numberOfLikes!! - 1
```
KlaKalma, line 29: Same here, null assert and safety check

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/ProfileViewModel.kt
```diff
@@ -229,6 +232,16 @@ constructor(
     _childrenPlaylistTrackList.value = (emptyList())
   }
 
+  fun likeTrack(track: Track) {
+    _wanderwaveLikedTracks.value += track
+    Log.d("ProfileViewModel", "Liked track: $track  " + _wanderwaveLikedTracks.value)
```
KlaKalma, line 237: Log

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/ProfileViewModel.kt
```diff
@@ -229,6 +232,16 @@ constructor(
...
+  }
+
+  fun unlikeTrack(track: Track) {
+    _wanderwaveLikedTracks.value -= track
+    Log.d("ProfileViewModel", "Uniked track: $track  " + _wanderwaveLikedTracks.value)
```
KlaKalma, line 242: Log

KlaKalma (changes requested):
Just some little comments nothing very bothering, I like the new like button

IB-12:
Thanks for the review, your inputs are greatly appreciated

app/src/main/java/ch/epfl/cs311/wanderwave/model/data/ProfileTrackAssociation.kt
```diff
@@ -20,13 +20,23 @@ data class ProfileTrackAssociation(
...
     return ProfileTrackAssociation(profile, track, likers + profile, likes + 1)
   }
 
   fun unlikeTrack(profile: Profile): ProfileTrackAssociation {
-    this.profile?.numberOfLikes = this.profile?.numberOfLikes!! - 1
+    if (!likers.contains(profile)) return this
+    if (this.profile != null) {
+      this.profile.numberOfLikes -= 1
+    }
     return ProfileTrackAssociation(profile, track, likers - profile, likes - 1)
```
KlaKalma, line 40: Sorry for bothering, maybe I'm missing something but wouldn't this return need to be in the if? Because if the profile already like this song we don't want to add it again to the likers no?

KlaKalma (changes requested):

app/src/main/java/ch/epfl/cs311/wanderwave/model/data/ProfileTrackAssociation.kt
IB-12: Yeah that's why there is `if (!likers.contains(profile)) return this` just above, which returns an unchanged ProfileTrackAssociation if the profile has already liked the song

KlaKalma: I meant adding null to the list, if profile is null we don't want to add it to the list no?

KlaKalma:
Perfect sir Thanks !

KlaKalma (approved):
lgtm

PR #362, opened 2024-05-23
add button to remove banned songs

### LINKED ISSUE #316: Users can remove songs from the banned songs list in the profile

joriba:
The "Banned songs" list on the TrackListScreen now has the capability to remove items from it. Closes #316.

The problem is that right now, the track list view is bugged, so it's a bit hard to verify correctness: the list just shows the liked songs instead of the banned ones. I created a preview composable that just shows the banlist, and you can check the animations by running the preview.

What remains to be done is to persist the changes, which means that removing a track from the list actually removes it from the list of banned songs on the profile. But the framework for that is not really there yet since there is no way of even getting the list of banned songs from the firebase profile, let alone modify it.


mbouaissi (approved):
lgtm, looks very cool ! Regarding the thing with that we can't test yet, i think that once Yoric stuff is merge, we can really test it

PR #384, opened 2024-05-23
Feature/like songs/firebase profile track assotiation

KlaKalma:
This PR requested by imade, implements the back end for the new track profile assotiation including ; 

- Bit of refactoring for the track profile assotiation (add the toMap and from functions)
- Bit of refactoring in the beaconConnection (use the to map 2 times and the from function)
- Implementation of the actual code in the trackConnection and BeaconConnection


IB-12 (approved):
LGTM, thank you

PR #391, opened 2024-05-24
Basic offline mode

IB-12:
This creates a basic offline mode where users can browse their tracklist, play songs if they are already downloaded, and view their profiles. 
Unfortunately, despite many many attempts, it seems like there is nothing in the Spotify API that lets us know which songs are available offline, which means that we can't filter them in the tracklist.


mbouaissi (approved):
lgtm

PR #395, opened 2024-05-24
Retrieve and drop songs auto

### LINKED ISSUE #383: Finish the thing about dropping songs and retrieving songs automatically

mbouaissi:
Now when we walk, this should drop a random song from the profile top songs, as well as recovering songs from beacon and putting them in the recently listen track. Now we can't really see it, but once all the bugfix are done, and the liked song integration with the profile is done, this should work. 

Closes #383


app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/MapScreen.kt
```diff
@@ -116,7 +116,8 @@ fun createLocationReceiver(
         }
         val tempBeacon = findClosestBeacon(locationState.value!!, viewModel.beaconList.value)
         if (tempBeacon != null) {
-          viewModel.getRandomSong(tempBeacon.id)
+          viewModel.getRandomSongAndAddToProfile(tempBeacon.id)
+          viewModel.retrieveSongFromProfileAndAddToBeacon(tempBeacon.id)
           Log.d(
               "MapScreen",
               "retrieved song: ${song.value}") // TODO: the song is retrieved here if exist
```
IB-12, line 123: What is this TODO about ? Make sure to remove it if has been addressed or to create an issue for it please

IB-12 (approved):
LGTM, you just forgot a TODO

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/MapViewModel.kt
```diff
@@ -43,6 +52,21 @@ constructor(val locationSource: LocationSource, private val beaconRepository: Be
...
+              lastName = "My LastName",
+              description = "My Description",
+              numberOfLikes = 0,
+              isPublic = true,
+              spotifyUid = "My Spotify UID",
+              firebaseUid = "My Firebase UID",
+              profilePictureUri = null))
+  val profile: StateFlow<Profile> = _profile
+  private var _profileUI = MutableStateFlow(ProfileViewModel.UIState())
+  val profileUI: StateFlow<ProfileViewModel.UIState> = _profileUI
```
joriba, line 68: Are the public states `profile` and `profileUI` necessary? From what I can see, `profileUI` is never used, and `profile` only in a test...

joriba: Also, I don't think both of the private states are necessary, right? UiState is just a profile with a loading and error parameter attached, but we don't need that here. I think you should be able to just use one flow.

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/MapScreen.kt
```diff
@@ -81,6 +81,7 @@ fun MapScreen(navigationActions: NavigationActions, viewModel: MapViewModel) {
     LaunchedEffect(true) {
       val intent = Intent(context, LocationUpdatesService::class.java)
       context.startService(intent)
+      viewModel.getProfileOfCurrentUser()
     }
 
     DisposableEffect(Unit) {
```
joriba, line 87: When does this get executed? Usually, when we have the map open, and walk around, we want to add and retrieve songs from beacons continuously, which I don't really see happening here. Can you clarify?

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/MapViewModel.kt
```diff
@@ -103,18 +127,59 @@ constructor(val locationSource: LocationSource, private val beaconRepository: Be
...
+          }
+          fetchedProfile.onFailure { exception ->
+            Log.e("Profile not found", "Profile not found for the given id $exception")
+          }
+        }
+      }
+    }
+  }
+
+  fun addTrackToBeacon(beaconId: String, track: Track, onComplete: (Boolean) -> Unit) {
```
joriba, line 164: Okay, this function is also almost identically there in the BeaconViewModel. Ideally, we'd want this logic to happen in the BeaconConnection, no?
But I understand if you feel that this refactoring is out of scope for your PR :)

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/MapViewModel.kt
```diff
@@ -103,18 +127,59 @@ constructor(val locationSource: LocationSource, private val beaconRepository: Be
...
         }
         fetchedBeacon.onFailure { exception ->
-          Log.e("No beacons found", "No beacons found for the given id")
+          Log.e("No beacons found", "No beacons found for the given id $exception")
+        }
+      }
+    }
+  }
+
+  fun retrieveSongFromProfileAndAddToBeacon(beaconId: String) {
```
joriba, line 146: Maybe make the fact that the song is chosen randomly reflect in the name of this function.

joriba (changes requested):

app/src/main/java/ch/epfl/cs311/wanderwave/ui/screens/MapScreen.kt
mbouaissi: This is executed at the launch of the application, once the user login we have the map, and then we retrieve the profile. To see where we add and drop the songs contiuously you can look below in the disposableEffect, we call createLocationReceiver, and here, we call the corresponding functions. This function was at the beginning looking for the POI near a user every 30seconds, but it can also do other things, as retrieving and droping songs. Also I thought that we only need to retrieve the user profile once(at the beginning), but maybe there is an other reason why we would want to fetch it more often?

app/src/main/java/ch/epfl/cs311/wanderwave/viewmodel/MapViewModel.kt
mbouaissi: I put the function in a new file in the utils folder

IB-12 (approved):
Still lgtm

PR #400, opened 2024-05-25
Change the theme

### LINKED ISSUE #399: modify the theme

mbouaissi:
This PR modify the main theme of the application. I think we should all discuss what we think is good and bad UI wise, so maybe we can do it in this PR.
For example, I didn't like the color yellow for the button, etc, and changes it in light orange. 
Same with the background, which was to dark, so I changed it in a dark blue one. 

Closes #399


IB-12 (approved):
Looks good ! Two things I wanted to mention:
1.  I think it'd be better if the logo was the same color as the accent color we're using
2.  Maybe we should try to go for a black background and use Spotify green as an accent ?

KlaKalma:
> Looks good ! Two things I wanted to mention:
> 
> 1. I think it'd be better if the logo was the same color as the accent color we're using
> 2. Maybe we should try to go for a black background and use Spotify green as an accent ?

For 1 I absolutely agree, I don't for 2, I think it's better if we have our own style. And I think that's what we decided at the begining no ? Yellow and red seemed fine.

IB-12 (approved):
LGTM

IB-12 (approved):
LGTM

PR #402, opened 2024-05-27
remove the zoom controls from the map screen

### LINKED ISSUE #375: [BUGFIX] remove the zoom controls (+, -) from the map screen

joriba:
Closes #375 

The "go to my location" button is still there, which I think is fine as this cannot be achieved via gestures, but let me know if I should remove it too.


KlaKalma (approved):
LGTM, I dont think we should remove the go to my location at all it's soo useful !

PR #403, opened 2024-05-27
Removed the "+" button from the "Recently Added" tracklist

IB-12:


aymanbakiri (approved):
Thanks for removing the + button from recently added tracklist.
LGTM

PR #404, opened 2024-05-27
370 bugfix bottom app bar is removed when going back to the beacon

### LINKED ISSUE #370: [BUGFIX] Bottom app bar is removed when going back to the beacon

joriba:
Fix the bug that made the bottom bar disappear. Closes #370 

The issue was that the route string that the navigationController gets is something like `beacon/something`, but the function that updates the route flow compares the strings by exact value, so only to `beacon`, which returns `null` for the Route.
Fixed by only comparing the top-level destination string, so everything up to the `/` in the route string.


IB-12 (approved):
LGTM 👍

PR #408, opened 2024-05-28
Fix severe issues

IB-12:
Get rid of the high severity issues on Sonarcloud (notably complexity in `App.kt`)


KlaKalma (approved):
lgtm, strange that just adding a comment is a severe issue but whatever.

PR #410, opened 2024-05-28
Add Wanderwave liked songs to a Spotify playlist

IB-12:
This creates a playlist in Spotify with the app's logo so that users can access their liked songs without needing to open Wanderwave. Didn't expect it to work, but somehow it does

Edit: If you want to try it out, you need to [unlink your Spotify account from Wanderwave](https://www.spotify.com/ch-de/account/apps/) and connect it again in order to enable the additional permissions required.


IB-12:
I don't have enough coverage yet, but I still think it'd be a valuable for the demo tomorrow, therefore I'll merge it as soon as it's reviewed

IB-12:
Thanks for the review, I'm as surprised as you are

PR #412, opened 2024-05-29
Fixed the issue with songs in beacons

IB-12:
The issue was caused by an incorrect argument in a function. I assume it was a merge error but it's fixed now


mbouaissi (approved):
lgtm, thx for solving this

PR #421, opened 2024-05-29
fix firebase problems

KlaKalma:


joriba (approved):
looks good from the code, why is the CI failing though?

PR #428, opened 2024-05-31
Tracks Id fixing

KlaKalma:
This PR Includes : 

- Fixing the Id of the tracks so they all have spotify:track: in front only one time
- Fixing the add track to ensure that the tracks cannot be duplicated and that the apps don't wipe the tracks list
- Using the AddTrackToBeacon when automatically pushing songs to beacons


mbouaissi (approved):
lgtm, and much nicer, thx for changing ths

mbouaissi (approved):
lgtm2

